PR #7, opened 2018-07-09
pahfit main code updates

alexmaragko:
Bellow is a list with the major changes:

1) Currently a new variable for the input spectrum is used, called "spectrum" (line 21). So the ArgumentParser is deactivated.
2) An example spectrum is uploaded to run pahfit: 'NGC7023-NW-BRIGHT.txt'
3) A drude profile function is implemented (line 83).
4) You can chose between drude and lorentzian profile. This is stored in the "fitprof" variable (line 34)
5) The pop-up plot is deactivated since it only shows the current best fit every "x" steps of iterations and shows no model components. In addition when the fit stops the fit in the window is not the best fit.
6) Plotting of all components and best fit is implemented. 
7) Residuals and best model are calculated.
8) The fit results are shown as a multi-line print instead of a 1-line print, and are saved into a file.
9) The saved fit results are then used to read and construct the multi-component plot.


karllark:
Closing this request.  Please, please add most of the code to the new code when it is appropriate.

PR #32, opened 2018-10-24
Update LICENSE

jdtsmith:
Switch to GPLv2 for continuity with `PAHFIT v1.x`.  Also need verbiage in each code file header ala:

```
  Copyright (C) 2018 PAHFIT Developers

  This file is part of PAHFIT.

  PAHFIT is free software; you can redistribute it and/or modify it
  under the terms of the GNU General Public License as published by
  the Free Software Foundation; either version 2, or (at your option)
  any later version.

  PAHFIT is distributed in the hope that it will be useful, but
  WITHOUT ANY WARRANTY; without even the implied warranty of
  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  General Public License for more details.

  You should have received a copy of the GNU General Public License
  along with PAHFIT; see the file LICENSE.  If not, write to the Free
  Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
  MA 02110-1301, USA.
```

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,29 +1,339 @@
-BSD 3-Clause License
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
 
-Copyright (c) 2017, Karl Gordon
-All rights reserved.
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
 
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are met:
+                            Preamble
 
-* Redistributions of source code must retain the above copyright notice, this
-  list of conditions and the following disclaimer.
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+License is intended to guarantee your freedom to share and change free
+software--to make sure the software is free for all its users.  This
+General Public License applies to most of the Free Software
+Foundation's software and to any other program whose authors commit to
+using it.  (Some other Free Software Foundation software is covered by
+the GNU Lesser General Public License instead.)  You can apply it to
+your programs, too.
 
-* Redistributions in binary form must reproduce the above copyright notice,
-  this list of conditions and the following disclaimer in the documentation
-  and/or other materials provided with the distribution.
+  When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+this service if you wish), that you receive source code or can get it
+if you want it, that you can change the software or use pieces of it
+in new free programs; and that you know you can do these things.
 
-* Neither the name of the copyright holder nor the names of its
-  contributors may be used to endorse or promote products derived from
-  this software without specific prior written permission.
+  To protect your rights, we need to make restrictions that forbid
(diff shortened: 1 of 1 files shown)


karllark:
@jdtsmith : can you specify why GPL over BSD for the license?  I ask know as an expert as all, but I've used BSD by default for all my python github projects, mainly as this is what astropy does.  Given that astropy picked BSD instead of GPL, I feel that a little more discussion on this decision would be good.  I do see the argument with going with the GPL license because that is what PAHFIT v1.2 used.

Lots of discussion online on this topic.  For example, simple googling brought me to:
https://www.freebsd.org/doc/en_US.ISO8859-1/articles/bsdl-gpl/article.html

karllark:
Why the need to have the license info in each code file?  The whole repository is covered by a single license file.  And the repository is published on github, with commits signed with a gpg key (verified and time stamped), so that should handle the legal stuff.  No great objection to having it in every file, expect the maintenance aspect (e.g., making changes in this header requires updating all headers).

jdtsmith:
In terms of file header, it's just a pointer, and can be minimized to copyright and see the LICENSE file.  The primary argument is that PAHFIT has been released under GPL2 for >12 years, so continuity is preferable.  The only other argument is that the BSD license allows 3rd party commercial software to integrate PAHFIT, make improvements, and sell their product without contributing back their improvements.  Is that likely?  Not at all.  But the GPL offers a stronger copyleft protection against that unfortunate if unlikely eventuality.

karllark:
You make strong arguments.  I have two examples of things we might want to do and I would like to understand if the GPL license would complicate things.

Example 1:
In creating the merged prototypes, I created a Drude1D class heavily based on the Lorentz1D class in astropy.modeling.  Not a problem as astropy is under the BSD license.  In the future, would it be straightforward to move the Drude1D class from PAHFIT to astropy.modeling?  The idea here is that others might find such a model useful.  My naive understanding of the GPL license is that this would not be possible without as the GPL license would require that astropy.modeling (at least) be under the GPL license.

Example 2:
The S07 extinction model embedded in the S07_attenuation class in the merged prototypes might be good to have in the more generic dust_extinction repository (that provides various extinction curve models for the community).  The dust_extinction code is under BSD.  Same question as above.

(So, really only 1 example).  :-)

jdtsmith:
The GPL and BSD 3-clause licenses are "compatible": https://www.gnu.org/licenses/license-list.en.html#ModifiedBSD.  By "compatible" they mean "you can combine code released under the other license with code released under the GNU GPL in one larger program."; see https://www.gnu.org/licenses/gpl-faq.html#WhatDoesCompatMean


Meaning in each of your examples, you could publish, say Drude1D under a BSD license and contribute it to astropy.modeling, which is not required to change its license.  You could also do this for any attenuation model.  PAHFIT can be readily distributed containing files licensed under both licenses.  What you may not do is to create a _derived_ version of PAHFIT, and redistribute it solely under a non-GPL license.

karllark:
Thanks for the explanation.  Much appreciated.  My understanding of these two licences has increased!

karllark:
I will add the info above to the prototype code in my pull request.  
Any objection to me merging this pull request?  That would represent the end of my review.  While we did not say it, I'm guessing the reviewer should not be the author of the pull request if at all possible in the case the author is one of the package maintainers.  ;-)

karllark:
BTW, did you want to update the README.rst file to have the right verbage about the license file?  Currently it states we use the BSD license.

karllark:
Superseded by another pull request (e.g., GPL v3 instead of GPL v2).

PR #45, opened 2019-01-29
Modelspec YML format draft, including PAHFIT classic

jdtsmith:
Here's a pull request for a draft version of the `.yml` files for both `PAHFIT` Classic, and a holding space for a more general format.  The latter in particular should achieve some flexibility in implementing constraints that combine parameters (e.g. `power("PAH7.7µm Complex")`, which pulls from multiple parameters of multiple features).  

In thinking about it, I don't think this needs to be an output format as well.  It could just be the entry format for specifying the model parameters, and the subsequent internal/on disk formats could be the simple flat table `parinfo` style.  The one drawback of this is that information on complex constraints is not propagated.


karllark:
Looks great to me.  Merging.

PR #49, opened 2019-02-12
rtd using pip and PEP8 fixes

karllark:
Fix RTD config files so that the automatic doc building should work.

Misc cleanup including removing a PDF file that was added incorrectly recently (missed it in the PR review).

Putting the absolute paths for module imports back.  This is better practice as it is how the package is used when installed on the system.  For developers, should use 'python setup.py develop'  to get the development code installed as if it had been done via pip/conda.


karllark:
@alexmaragko : note the return to absolute paths for imports.  Recommended by PEP8 (https://www.python.org/dev/peps/pep-0008/).  I find these easier to understand and they generally are more robust than relative imports.

karllark:
Merging as this is the only way to check the RTD fix actually fixed things!

Issue #58, opened 2019-03-22
Error when reading input file and unit conversion


alexmaragko:
In the current version of the code the input file format is decided from the input file extension.
If I run pahfit for the example spectra in the data directory it raises an error of "No reader defined for format ...", followed by a table with the available formats.

This is strange because .ipac is included in the available formats, but "M101_Nucleus_irs.ipac" gives the above error.

The same happens with "orion_bar_SWS_with7023shape.txt", although if I rename the file to .ascii it will read the file.

But, it will raise an error in the following line when the unit conversion is happening, because no unit assignment is happening.

So currently we can't run pahfit as: run_pahfit spectrumfile scipack_ExGal_SpitzerIRSSLLL.ipac without modifying the code, and it's harder to test.

karllark:
@alexmaragko : have you installed pahfit?  This is now needed to get the run_pahfit command installed on your system.  Check out the install instructions at:
https://pahfit.readthedocs.io/en/latest/install.html

I think this will solve the problem, but am not sure.  Please try and tell me.  :-)

alexmaragko:
@karllark Yes I have installed pahfit.

alexmaragko:
I will modify the code a bit to get the units in and make some runs for Monday's telecon though. :)

karllark:
Can you post the output error message?  I don't understand why it doesn't work if you have installed it.

alexmaragko:
This is the output for M101_Nucleus_irs.ipac:
Traceback (most recent call last):
  File "/usr/local/bin/run_pahfit", line 4, in <module>
    __import__('pkg_resources').run_script('pahfit==2.0.dev88', 'run_pahfit')
  File "/usr/local/lib/python3.5/dist-packages/pkg_resources/__init__.py", line 750, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/local/lib/python3.5/dist-packages/pkg_resources/__init__.py", line 1527, in run_script
    exec(code, namespace, namespace)
  File "/usr/local/lib/python3.5/dist-packages/pahfit-2.0.dev88-py3.5.egg/EGG-INFO/scripts/run_pahfit", line 88, in <module>
    pmodel = PAHFITBase(filename=packfile)
  File "/usr/local/lib/python3.5/dist-packages/pahfit-2.0.dev88-py3.5.egg/pahfit/base.py", line 123, in __init__
    param_info = self.read(filename, tformat=tformat)
[...]

PR #61, opened 2019-03-26
split index fix

alexmaragko:
This PR fixes the split index bug. Now pahfit will run for all ipac input spectra without errors. However, txt/ascii formats will still raise an error due to either the 'txt' extension or absence of units in the header.


karllark:
Great.  Merging.

karllark:
Partially addresses #58.

PR #62, opened 2019-03-30
Fix data format

alexmaragko:
Replacing ascii spectum with ipac. Including new ecsv spectrum. Applying fixes to plot_pahfit script. Should fully address issue #58.


karllark:
Great!  Merging.

PR #67, opened 2020-05-20
BlackBody1D/Drude1D classes removed and called from astropy

alexmaragko:
Removed BlackBody1D and Drude1D classes from component_models, and calling now the corresponding classes from astropy. Updated base.py.


karllark:
You should definitely not make the change for the astropy.modeling BlackBody in this pull request.  After I played around with this a bit, I queried one of the astropy.modeling core developers and it looks like a change to astropy is needed.  Fortunately, they are working on it already!!!!  
See https://github.com/astropy/astropy/pull/10415.  And I'm testing the change.  

So, we should do the change to BlackBody and units separate as it will require extra updates for sure!

karllark:
Superseded by #68.

PR #73, opened 2020-08-27
Fix glitch at 12 um in silicate attenuation curve

ThomasSYLai:
The original silicate attenuation curve was not well stitched at ~12 um. This pull request fixes the issue. 

Screen Shot 2020-08-27 at 2 49 13 PM


karllark:
Most excellent!  Is this PR ready to merge?   Looks like it is, just checking.

ThomasSYLai:
Yes, it should be ready.

PR #80, opened 2021-01-11
Upload AKARI scipack (no attenuation yet) and change to SpitzerIRSSLLL

ThomasSYLai:
Here is the newly uploaded AKARI science pack and some modifications to SpitzerIRSSLLL.ipac.
Note that the absorption features, e.g., 3.05 ice feature, CO2 (4.27 um), and CO (4.67 um), are not included in this update. 

Here is an example of fitting the 1C spectrum in Lai et al. (2020). Note that 1C is a composite spectrum of star-forming galaxies, with only moderate Si absorption. 
Screen Shot 2021-01-11 at 4 33 20 PM


ThomasSYLai:
This pull request consists of multiple issues. Need to be separated.

Issue #75, opened 2021-01-11
Attenuation curve flexibility 


ThomasSYLai:
The assumed attenuation curve is now fixed. A more flexible model for handling attenuation should be added. For example, a 3.1 um ice feature should be added on top of the silicate extinction curve. Any ideas?

karllark:
This will definitely be needed for data below 5 micron.
Maybe wish to treat the dust attenuation curve like the other components - in other words have it build from components.  I could see this as having components for the continuum, the silicate features, the ice features, etc.  And either using tables for the continuum/features or having analytic functions (if such exist - hint working on for the silicate features myself).

PR #83, opened 2021-01-12
Upload AKARI scipack (no attenuation yet)

ThomasSYLai:
Here is the newly uploaded AKARI science pack, which includes additional emission features from 2.6--5 um as listed (in unit of um).

Dust features:
PAH - 3.29
Aliphatic - 3.4
Plateau - 3.47
p.s. PAH 5.27 is split into two features at 5.24 and 5.33.

Br beta - 2.63
Pf gamma - 3.74
Br alpha - 4.05
Pf beta - 4.65

Note that the absorption features, e.g., 3.05 and 6.02 ice feature, CO2 (4.27), and CO (4.67), are not included in this update. Further updates to the issues on how to deal with attenuation will be in #75.

The following figure shows the fitted 1C spectrum in Lai et al. (2020). Note that 1C is a composite spectrum of star-forming galaxies, with only moderate H2O ice and Si absorption. So, the fitter can present a reasonable fit in between 2.5-5 um without the need for attenuation, but for more obscured sources, incorporating proper attenuation profiles is necessary.

Screen Shot 2021-01-11 at 9 56 04 PM

Lastly, given that x_0 and FWHM are fixed in this update, the min and max are crudely set and should not affect the result.


karllark:
This is great.  Would it be possible to include example data that includes AKARI+Spitzer data?  Like something from your thesis paper?

karllark:
Should this PR be closed yet?

Issue #82, opened 2021-01-11
Parameter Starting Point


jdtsmith:
In addition to science pack components (positions, widths, etc. for all the features), the final fit is impacted by the _starting position_ of the fit in the parameter space.  In IDL PAHFIT, this starting point was crudely determined for each fitted spectrum:

- **Stellar Blackbody**: The stellar blackbody (T=5000K) was evaluated at the shortest wavelength (at least 5.25µm), and divided into the actual intensity being fitted at that wavelength. This sets the starting `tau_star` value.  

- **Thermal Dust Continuum**: Each modify blackbody `n I_nu` is evaluated at it's maximum wavelength, and divided into the interpolated intensity at this wavelength.  This sets the `tau_dust_*` parameters. 

- **Lines and Dust Emission Features (PAHs)**: The central intensity for all lines is set to 1/5th of the median (non-negative) intensity of the entire input spectrum.   

- **Attenuation/Silicate opacity**: A constant tau_Si=0.1 was adopted.  

These aren't necessarily optimal (especially estimating tau_Si and getting _in the ballpark_ would be valuable to the fitter's performance and quality of its results).  

But per-spectra starting point estimation increases the robustness of fits, and will make IDL PAHFIT comparable with classic pack for automated testing.  This is important enough to live in its own module I think.

Since this affects the model to some degree, perhaps a science pack can optionally specify which "parameterStarter" it wants to utilize (with the default being like the above, but no doubt with improvements coming).

This likely needs to be part of the first release to be usable on archival spectra.

karllark:
Shouldn't this be in the v2.0 release?  As this is required to do the basic IDL vs python comparison.
Good to keep the v2.1 release as just about AKARI I think.

jdtsmith:
Reasonable.  Probably should move the due date of 2.0 back.

jdtsmith:
Current "packs":

**Instrument Pack**: information about instrumental parameters, such as wavelength range and cutoff, line spread function (as a function of wavelength), etc.

**Science Pack**: Scientifically focused parameter setup, including constraints, total number of features, exploring the tradeoff between fit degeneracy and flexibility.  

[...]

PR #92, opened 2021-01-26
Add starting point

ThomasSYLai:
Adding the function -- starting_point in base.py, which takes the input spectrum and guesses the starting points of the components' amplitude, based on the suggestions in #82. Note that this function will not be applied if amp_fixed in each component is set to be True in the science pack file.

### DIFF
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -5,6 +5,8 @@ from pahfit.component_models import BlackBody1D, S07_attenuation
 from astropy.table import Table, vstack
 from astropy.modeling.physical_models import Drude1D
 
+from scipy import interpolate
+
 import numpy as np
 
 __all__ = ["PAHFITBase"]
@@ -105,7 +107,7 @@ class PAHFITBase:
         x_0, x_0_limits, x_0_fixed, fwhms, fwhms_limits, fwhm_fixed}.
     """
 
-    def __init__(self, param_info=None, filename=None, tformat=None):
+    def __init__(self, obs_x, obs_y, param_info=None, filename=None, tformat=None):
         """
         Setup a variant based on inputs.  Generates an astropy.modeling
         compound model.
@@ -120,6 +122,9 @@ class PAHFITBase:
         # read in the parameter info from a file
         if filename is not None:
             param_info = self.read(filename, tformat=tformat)
+
+        # guess values and update starting point (if not set fixed) based on the input spectrum
+        param_info = self.starting_point(obs_x, obs_y, param_info)
 
         bb_info = param_info[0]
         dust_features = param_info[1]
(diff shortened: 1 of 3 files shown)


karllark:
Update needed to the pahfit/tests/test_classic_pack.py

See https://github.com/PAHFIT/pahfit/pull/92/checks?check_run_id=1771328090  (scroll down to find the error message).

karllark:
Also, codestyle error.  See https://github.com/PAHFIT/pahfit/pull/92/checks?check_run_id=1771327854

karllark:
Great to see this PR.  Excited to see how this changes the fitting.

ThomasSYLai:
@karllark There's one check that did not go through -- Codacy Static Code Analysis. Not sure if this is something we should be worried about.

pahfit/base.py
```diff
@@ -564,3 +569,69 @@ def read(self, filename, tformat=None):
         readout = (bb_info, df_info, h2_info, ion_info, att_info)
 
         return readout
+
+    def starting_point(self, obs_x, obs_y, param_info):
```
karllark, line 573: The codacy "error" is that this function does not use self, so it does not need to be in this class.  It is recommending a stand alone function as a result.  Otherwise, you could declare this function as a static.  See https://realpython.com/instance-class-and-static-methods-demystified/

pahfit/base.py
```diff
@@ -564,3 +569,69 @@ def read(self, filename, tformat=None):
...
+        """
+        return updated starting point in param_info based on the input spectrum
```
karllark, line 575: This is a very minimal docstring.  It would be great if this could be expanded to the more standard format like that at the very start of the class.

pahfit/base.py
```diff
@@ -564,3 +569,69 @@ def read(self, filename, tformat=None):
...
+        """
+
+        # guess starting point of bb
+        sp = interpolate.splrep(obs_x, obs_y)
+        for i, (fix, temp) in enumerate(zip(param_info[0]['amps_fixed'], param_info[0]['temps'])):
+
+            if (fix is False) & (i == 0):  # BB0 - stellar comoponent measured at 5.5 um
```
karllark, line 582: There is no guarantee that the stellar BB component will be at the beginning, no?  Or am I wrong?  If not, then hopefully there is a way to identify the stellar component other than i==0.

pahfit/tests/test_classic_pack.py
```diff
@@ -17,9 +19,20 @@ def test_classic_pack():
...
+
+    obs_spectrum = Table.read(datafilename, format='ipac')
+    obs_x = obs_spectrum['wavelength'].to(u.micron,
+                                          equivalencies=u.spectral())
+    obs_y = obs_spectrum['flux'].to(u.Jy,
+                                    equivalencies=u.spectral_density(obs_x))
+    obs_x = obs_x.value
+    obs_y = obs_y.value
+
+    pmodel = PAHFITBase(obs_x, obs_y, filename=packfilename, tformat='ipac')
```
karllark, line 35: Why not just use obs_x.value and obs_y.value in the call to PAHFITBase?  Then you don't need lines 32-33.

karllark (changes requested):
Looks good.  A few comments.

pahfit/base.py
ThomasSYLai: I've changed it to a staticmethod.

pahfit/base.py
ThomasSYLai: I've defined the stellar component to be BB that has temperature >= 2500K. We can perhaps change it to other temp if needed.

pahfit/base.py
ThomasSYLai: Docstring has been added.

PR #93, opened 2021-02-08
Fit parameters to match IDL mpfitfun

### LINKED ISSUE #84: Setup fitter to be have similar parameters values to IDL version
The parameters for the fitter need to be setup to be similar to the IDL version.  Specifically maximum number of interations, convergence criteria, etc.  This is to take advantage of the extensive testing the IDL version has undergone and to provide a fair comparison between the versions.

Will impact #74.

karllark:
For v2.0, the goal is to have the python version to be very similar to the IDL version.  One area for this is the parameters that impact the quality and speed of the fit.  Read the doc for [mpfitfun.pro](https://pages.physics.wisc.edu/~craigm/idl/down/mpfitfun.pro) and the docs for the astropy [LevMarLSQFitter](https://docs.astropy.org/en/stable/api/astropy.modeling.fitting.LevMarLSQFitter.html) and the called the scipy [leastsq](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.leastsq.html#scipy.optimize.leastsq).  To my understanding the max number of iterations (200) and tolerances (1e-10) needed to be set.

Closes #84.

### DIFF
--- a/scripts/run_pahfit
+++ b/scripts/run_pahfit
@@ -94,7 +94,7 @@ if __name__ == '__main__':
 
     # fit
     obs_fit = fit(pmodel.model, obs_x, obs_y, weights=weights,
-                  maxiter=1000)
+                  maxiter=200, epsilon=1e-10, acc=1e-10)
     print(fit.fit_info['message'])
 
     # save results to fits file


jdtsmith:
Looks sensible to me.  It does re-raise the question of whether fitter details (including which optimizer to use) is a free and separate choice from a science pack.  Maybe not, since some optimizers allow expressing more complicated constraints.  Possibly we should hide all those details from the user.

jdtsmith (approved):
Looks sensible to me. It does re-raise the question of whether fitter details (including which optimizer to use) is a free and separate choice from a science pack. Maybe not, since some optimizers allow expressing more complicated constraints. Possibly we should hide all those details from the user.

karllark:
BTW, with these changes fitting the M101 nucleus takes ~50 seconds on my (not very fast) laptop.  Do not have IDL on my laptop, so can't do the direct IDL timing test.

karllark:
> It does re-raise the question of whether fitter details (including which optimizer to use) is a free and separate choice from a science pack. Maybe not, since some optimizers allow expressing more complicated constraints. Possibly we should hide all those details from the user.

Yep.  I'm going to punt this to post-v2.0 work.  Focusing for now on getting v2.0 done.

PR #94, opened 2021-02-08
Uplate scipack_ExGal_SpitzerIRSSLLL to make it identical to IDL parinfo

ThomasSYLai:
Here are the updated parameters for scipackExGal_SpitzerIRSSLLL directly translated from the input parameters of IDL parinfo. The main changes include setting the central wavelength and fwhm of DF as fixed and setting the starting point of tau_Si from 1 to 0.1 with no upper bound. The amp of BB, DF, and lines are guessed initial values from inputting the M101_nucleus file in the IDL PAHFIT.


karllark (changes requested):
The test_classic_pack is failing.  Please update the test to be consistent with your changes to the pack file.  This is the simple test that the classic pack on disk matches the classic pack defined in PAHFIT_Spitzer_Exgal.py.

els1:
Conform IDL settings. Checks need to resolved.

PR #95, opened 2021-02-08
Small changes in the  factors of starting point in DF, H2, and ion

ThomasSYLai:
Fix typos in the factors of calculating the starting points of DF, H2, and ion.


karllark (approved):
Looks great.

PR #100, opened 2021-03-05
Noguess

alexmaragko:
Added a "noguess" flag. By default the guess function will be used, unless the "noguess" flag is given. The "noguess" flag is set True by default in plot_pahfit.

### DIFF
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -126,8 +126,9 @@ class PAHFITBase:
         if filename is not None:
             param_info = self.read(filename, tformat=tformat)
 
-        # guess values and update starting point (if not set fixed) based on the input spectrum
-        param_info = self.guess(obs_x, obs_y, param_info)
+        if not noguess:
+            # guess values and update starting point (if not set fixed) based on the input spectrum
+            param_info = self.guess(obs_x, obs_y, param_info)
 
         bb_info = param_info[0]
         dust_features = param_info[1]
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -110,7 +110,7 @@ class PAHFITBase:
         x_0, x_0_limits, x_0_fixed, fwhms, fwhms_limits, fwhm_fixed}.
     """
 
-    def __init__(self, obs_x, obs_y, param_info=None, filename=None, tformat=None):
+    def __init__(self, obs_x, obs_y, noguess, param_info=None, filename=None, tformat=None):
         """
         Setup a variant based on inputs.  Generates an astropy.modeling
         compound model.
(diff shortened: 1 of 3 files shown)


pahfit/base.py
```diff
@@ -110,7 +110,7 @@ class PAHFITBase:
         x_0, x_0_limits, x_0_fixed, fwhms, fwhms_limits, fwhm_fixed}.
     """
 
-    def __init__(self, obs_x, obs_y, param_info=None, filename=None, tformat=None):
+    def __init__(self, obs_x, obs_y, noguess, param_info=None, filename=None, tformat=None):
```
karllark, line 113: This should be changed to a boolean keyword argument.  Or do you think it is a required argument and you want to the users to explicitly input it.

For a keyword argument, I would suggest it should be renamed to "estimatestart" or "estimateinnit" something like that.  And then have it only call the "guess" function if it is set to True.  This would mean that the default behavior is to not do anything and I think this makes more sense when thinking about instantiating a class.  At least in my brain.

scripts/run_pahfit
```diff
@@ -37,7 +37,9 @@ def initialize_parser():
     parser.add_argument('--saveoutput', action='store',
                         default='ipac', choices=savetypes,
                         help='Save fit results to a file of specified type')
-
+    parser.add_argument('--noguess', action='store_true',
```
karllark, line 40: I don't find "guess" as a very intuitive descriptor.  Like my comment above, I like something like estimate.  With the description text list "Estimate initial starting fit parameter values".

karllark (changes requested):
Great start.  A few modifications suggested.

karllark:
Maybe rename the "guess" function as part of this PR to have a more descriptive name.  Like "estimate_initital" or "est_init_param" or "est_init_from_obs" or something like that.

karllark:
The tests are failing as the positional "noguess" parameter is missing in at least one of the test functions.

pahfit/base.py
alexmaragko: I was thinking it might be useful to set it as a user defined keyword argument, because people might want to completely bypass the guess function, for instance if re-fitting a spectrum and you want to use as a starting point the initial run results and not go through the guess function again. 

No preference on the naming. "estimatestart" or "estimateinit" seems good to me.

karllark: Currently it is not a keyword.  But a required positional argument.  To set it as a keyword, it would have to be "noguess=True" for example.

alexmaragko: Sorry I was thinking about a positional argument, because currently there is no other "interaction" with the user other than providing arguments.

scripts/run_pahfit
alexmaragko: I used the word "guess",  because "guess" is what we currently call the method in the code. But I have no preference one the naming as I mentioned above. :)

alexmaragko: I can rename the "guess" function as well as you proposed in the comment below.

karllark: Great!

alexmaragko:
Will make new PR

PR #101, opened 2021-03-09
Estimate init

alexmaragko:
- Renamed guess method to estimate_init
- Renamed the positional argument to noestimate in run_pahfit
- Added noestimate=False argument  in test_classic_pack.py and set to False in plot_pahfit

### DIFF
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -110,7 +110,7 @@ class PAHFITBase:
         x_0, x_0_limits, x_0_fixed, fwhms, fwhms_limits, fwhm_fixed}.
     """
 
-    def __init__(self, obs_x, obs_y, estimate_init, param_info=None, filename=None, tformat=None):
+    def __init__(self, obs_x, obs_y, noestimate, param_info=None, filename=None, tformat=None):
         """
         Setup a variant based on inputs.  Generates an astropy.modeling
         compound model.
@@ -126,7 +126,7 @@ class PAHFITBase:
         if filename is not None:
             param_info = self.read(filename, tformat=tformat)
 
-        if estimate_init:
+        if not noestimate:
             # guess values and update starting point (if not set fixed) based on the input spectrum
             param_info = self.estimate_init(obs_x, obs_y, param_info)
 
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -128,7 +128,7 @@ class PAHFITBase:
 
         if estimate_init:
             # guess values and update starting point (if not set fixed) based on the input spectrum
-            param_info = self.guess(obs_x, obs_y, param_info)
+            param_info = self.estimate_init(obs_x, obs_y, param_info)
 
         bb_info = param_info[0]
         dust_features = param_info[1]
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -110,7 +110,7 @@ class PAHFITBase:
         x_0, x_0_limits, x_0_fixed, fwhms, fwhms_limits, fwhm_fixed}.
     """
 
-    def __init__(self, obs_x, obs_y, param_info=None, filename=None, tformat=None):
+    def __init__(self, obs_x, obs_y, estimate_init, param_info=None, filename=None, tformat=None):
         """
         Setup a variant based on inputs.  Generates an astropy.modeling
         compound model.
@@ -126,8 +126,9 @@ class PAHFITBase:
         if filename is not None:
             param_info = self.read(filename, tformat=tformat)
 
-        # guess values and update starting point (if not set fixed) based on the input spectrum
-        param_info = self.guess(obs_x, obs_y, param_info)
+        if estimate_init:
+            # guess values and update starting point (if not set fixed) based on the input spectrum
+            param_info = self.guess(obs_x, obs_y, param_info)
 
         bb_info = param_info[0]
         dust_features = param_info[1]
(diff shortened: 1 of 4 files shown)


pahfit/base.py
```diff
@@ -110,7 +110,7 @@ class PAHFITBase:
         x_0, x_0_limits, x_0_fixed, fwhms, fwhms_limits, fwhm_fixed}.
     """
 
-    def __init__(self, obs_x, obs_y, param_info=None, filename=None, tformat=None):
+    def __init__(self, obs_x, obs_y, noestimate, param_info=None, filename=None, tformat=None):
```
karllark, line 113: I find it a bit confusing to have a parameter that a negative.  What would you think of making this parameter name `estimate_start` so that it is a positive and a bit more descriptive?  And if you made it `estimate_start=False` then the default behavior would that only `obs_x` and `obs_y` were needed when defining the `PAHFITBase` class.  As we will use this class for more than fitting, the default then would be use the pack file values and this seems more intuitive to me.

karllark (changes requested):
Looking better.

BTW, no need to open a new PR when making changes.  Just push changes to the same branch and the PR will be automatically updated.  This is preferred as it then keeps the full history of the PR.

pahfit/base.py
karllark: > I made a new branch and PR because my previous branch was 1 commit behind after last night's #94 merge, and I wasn't sure how it would behave. :)

It likely would have been fine.  And there are ways to solve this if it wan't.  Ok to open a new PR, just wanted to check on the reason.

karllark: We need to make sure that calling `PAHFITBase` works both from the scripts and just calling (for example) `a = PAHFITBase(obs_x, obs_y)` works.

karllark:
Please push any changes to this branch.  A PR was just merged that should mean the tests will now pass.  And shouldn't conflict at all with this PR.

pahfit/base.py
alexmaragko: Sorry I'm not following completely how you'd want this to be implemented. Would you prefer it to be:

1) a user provided argument, e.g.  run_pahfit M101_Nucleus_irs.ipac scipack_ExGal_SpitzerIRSSLLL.ipac --newargument

if so, do you want it to enable the estimate_init or bypass it? This will determine both the naming (e.g. estimate_start, or estimate_skip respectively) and the true/false boolean value stored, and subsequently the condition in base.py for using the estimate_init.

2) a keyword argument in the run_pahfit script that the user will need to edit run_pahfit to provide it.

3) something else.

My thoughts on this is that the user will only have to bother when they want to bypass the estimation of the initial conditions. And the easiest way to ask a user to do it is to provide  a --newargument than editing a script file.

Please let me know if I misunderstood something in your comments.

alexmaragko: @karllark any comments on the above?

karllark: Will put in comments tomorrow.  Been distracted.  :-)

karllark: We should have 1 and 2.  This supports using PAHFITBase from the command line using a script (like `run_pahfit`) and from code/notebooks by direct python call (using `a = PAHFITBase(obs_x, obs_y)).

I feel the use should be the same in both cases - a keyword that enables using the estimate_init code.  The default should be to *not* use the estimate_init code to support the broadest use of PAHFITBase.  It would be clearer if the keyword for the script and PAHFITBase would be the same.  And `estimate_start` sounds the best to me for this keyword as it best conveys that the *starting* point of the fit is *estimated*.

To be explicit as code can be easier to understand the call for the PAHFIT init function would be:
`def __init__(self, obs_x, obs_y, estimate_start=False, param_info=None, filename=None, tformat=None):`

and the argparse argument would be:
`parser.add_argument('--estimate_start', action='store_true', help='Estimate of starting point based on the input spectrum')`
And this means that `args.estimate_start = False` unless the user specifies `--estimate_start` on the commandline.

If this is not clear, please say so.  Happy to have a quick video call to discuss this if desired.

Various others parts of the code would need to be updated of course.

alexmaragko: This is clear. We also need to update documentation regarding the usage of estimate_start.

alexmaragko: Changes made. I tested it locally with and without the argparse argument and behaves as expected.

pahfit/tests/test_classic_pack.py
```diff
@@ -30,7 +30,7 @@ def test_classic_pack():
     obs_y = obs_spectrum['flux'].to(u.Jy,
                                     equivalencies=u.spectral_density(obs_x))
 
-    pmodel = PAHFITBase(obs_x.value, obs_y.value, filename=packfilename, tformat='ipac')
+    pmodel = PAHFITBase(obs_x.value, obs_y.value, estimate_start=False, filename=packfilename, tformat='ipac')
```
karllark, line 33: Since `estimate_start=False` is the default, you do not need to put include this in the PAHFITBase call.

scripts/plot_pahfit
```diff
@@ -70,7 +70,7 @@ if __name__ == "__main__":
     obs_y = obs_y.value
 
     # read in the PAHFIT results
-    pmodel = PAHFITBase(obs_x, obs_y, filename=args.fitfilename)
+    pmodel = PAHFITBase(obs_x, obs_y, estimate_start=False, filename=args.fitfilename)
```
karllark, line 73: Since `estimate_start=False` is the default, you do not need to put include this in the PAHFITBase call.

karllark (commented):
Looks great.  A minor comment.

scripts/plot_pahfit
alexmaragko: Got it, thanks.

karllark (approved):
Looks great.  Thanks for responding to all my comments - very much appreciated.

alexmaragko:
Thank you for the constructive comments, I sorted out a couple of things on arguments.

PR #102, opened 2021-03-09
Fix test that our "classic" pack data file is correct

karllark:
Required updating the software version of the classic pack that is used to check the pack file.  Also update the pack file to go back to the "old" default values of the amplitudes as these are arbitrary and this avoided having to update the software code.

I also updated the classic pack file to be more human readable by removing the extraneous digits.  Also makes the width of the table significantly smaller.


karllark:
> How did you decide on the limits for the FWHM of the Drude profiles?

I used the commented out code from the IDL version that sets them as +/- 10% of the initial values.

Issue #47, opened 2019-02-11
YAML format reader


karllark:
The proposed YAML format needs a reader and appropriate integration into the class init function.  Probably would be good to have the reader picked based on the file extension (e.g., switching between the YAML and straight table formats).

jdtsmith:
The optimizer should be part of the science pack YAML format.

PR #107, opened 2021-03-16
Scripts now callable system wide after pip install

### LINKED ISSUE #106: Docs: example in documentation not working
The examples to run and plot are not working in the documentation.  This is because the scripts `run_pahfit` and `plot_pahfit` are not being installed system wide as they used to be after pip installing pahfit.  This is a bug and emerged from the update to the new packaging files.  Need to change the `run_pahfit` and `fit_pahfit` to be entry points.

karllark:
Make the `run_pahfit` and `plot_pahfit` scripts callable system wide after pip install.  This uses entry-points.  This restores the previous behavior that had these scripts installed system wide using a different method.  Entry-points are more pythonic (I think) and supported with the new packaging files.  This means our current docs for the example running and plotting of results are now correct again.  Locations of the scripts moved to inside the pahfit directory to allow entry-points to work.

Closes #106.

### DIFF
--- a/pahfit/scripts/plot_pahfit.py
+++ b/pahfit/scripts/plot_pahfit.py
@@ -14,22 +14,37 @@ from pahfit.base import PAHFITBase
 
 
 def initialize_parser():
-    """
-    Command line parser for plot_PAHFIT
+    """Command line parser for plot_pahfit
 
     Returns
     -------
     parser : argparse object
     """
-    plottypes = ['png', 'jpg', 'jpeg', 'pdf', 'ps', 'eps', 'rgba',
-                 'svg', 'tiff', 'tif', 'pgf', 'svgz', 'raw']
+    plottypes = [
+        "png",
+        "jpg",
+        "jpeg",
+        "pdf",
+        "ps",
+        "eps",
+        "rgba",
+        "svg",
+        "tiff",
+        "tif",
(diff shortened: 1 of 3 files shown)


jdtsmith:
I haven't tested this, but it looks reasonable to restore prior behavior.  That said, one very good goal is to enable PAHFIT to work well & seamlessly in interactive sessions such as Jupyter Notebooks or iPython.  This will require moving some/most of the "accept a spectrum Table", "setup the optimizer" etc. inside the base or an interface class itself.  So users can try out various science packs on the same spectrum, etc.  Also see #47, which is relevant.

karllark:
> That said, one very good goal is to enable PAHFIT to work well & seamlessly in interactive sessions such as Jupyter Notebooks or iPython. This will require moving some/most of the "accept a spectrum Table", "setup the optimizer" etc. inside the base or an interface class itself.

Can you make an issue for this?  One solution would be to provide examples of how to do this.  We definitely should support different use cases for pahfit including commandline, python code, notebooks, interactive python sessions, ...

karllark:
@jdtsmith : if you found this PR ok - please formally approve.

jdtsmith (approved):
Looks good.

karllark:
Thanks.  Will merge.

PR #111, opened 2021-03-24
Indentation and param_Info placement fix

alexmaragko:
Fixed the placement of param_info[x]['amps'][i] = amp_guess for BBs and its indentation for the rest of the parameters. Now the estimate function will work only for non-fixed amplitudes.


els1 (approved):
Looks good.

PR #113, opened 2021-04-05
Add fitting test (and refactor scripts)

karllark:
Add in a fitting test.  Simply reads in the spitzer scipack and M101 nucleus spectrum, runs the fit, and then compares the fit parameters to a cached array of the parameters.  Will fail if the parameters are not close (machine precision I think).

Refactored run_pahfit and plot_pahfit scripts to have common functions to avoid code duplication and allow for easier testing.  This refactoring will also make it easier to support other fitters, units, etc. in the future.


pahfit/helpers.py
```diff
@@ -0,0 +1,138 @@
...
+    tformat = specfile.split(".")[-1]
+    if tformat == "ecsv":
+        tformat = "ascii.ecsv"
+    obs_spectrum = Table.read(specfile, format=tformat)
+    obsdata = {}
+    obsdata["x"] = obs_spectrum["wavelength"].to(u.micron, equivalencies=u.spectral())
+    obsdata["y"] = obs_spectrum["flux"].to(
+        u.Jy, equivalencies=u.spectral_density(obsdata["x"])
+    )
+    obsdata["unc"] = obs_spectrum["sigma"].to(
```
alexmaragko, line 51: By using "wavelength", "flux" and "sigma" here we imply that the input spectrum's must have the exact column naming in the header, otherwise it will raise an error. Perhaps we can allow more freedom by retrieving the data column names as provided by the user, using colnames (colnames[0], colnames[1], and colnames[2)]. This way we don't force a naming convention, instead we only request to have the wavelength, flux, and uncertainty columns in that order, and we can specify it in the docs.

alexmaragko (commented):
Looks good! Perhaps consider some of the changes I propose in my comments, otherwise it seems good as it is, and definitely looks much cleaner now.

pahfit/helpers.py
```diff
@@ -0,0 +1,138 @@
...
+    Returns
+    -------
+    obsdata : dict
+        x is wavelength in microns and y/unc are the spectrum/unc in units
+        of Jy
+    """
+
+    # read in the observed spectrum
+    # assumed to be astropy table compatibile and include units
+    specfile = specfile
```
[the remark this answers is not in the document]
karllark: Good point!  Hah.

pahfit/helpers.py
karllark: Good idea.  Will implement as a keyword to this function with the current as the default.

els1 (commented):
Looks good to me overall.  
One question: for 'test_fitting_spitzer.py', you use maxiter=200 (ln 17) while the default is set to 1000. Is this to decrease the time for testing or is this a leftover from the previous setting?
And how do I show that part of the code in this comment (like Alexandros did)?

karllark:
> One question: for 'test_fitting_spitzer.py', you use maxiter=200 (ln 17) while the default is set to 1000. Is this to decrease the time for testing or is this a leftover from the previous setting?

For testing so it is fully reproducible regardless of the default maxiter setting.  And to reduce the time needed.

karllark:
> And how do I show that part of the code in this comment (like Alexandros did)?

If you go to the "Files changed" tab, you can highlight specific lines and add a comment/review.  Use your mouse and the "+" symbols.

els1:
> > And how do I show that part of the code in this comment (like Alexandros did)?
> 
> If you go to the "Files changed" tab, you can highlight specific lines and add a comment/review. Use your mouse and the "+" symbols.

Thanks.

karllark:
Two reviews enough?  I think it is ready to merge.

els1:
> Two reviews enough? I think it is ready to merge.

Fine with me. But there is one check which was unsuccessful ...

karllark:
> But there is one check which was unsuccessful ...

This is codacy being over careful in the default value for a variable in a function definition.   It can be ignored.

PR #114, opened 2021-04-19
Add axis of attenuation in the output plot

ThomasSYLai:
Add an axis of attenuation at the right of the output plot to characterize the amount of attenuation.

image


els1 (approved):
Looks good.

PR #115, opened 2021-04-19
Update plot

alexmaragko:
Update plot function in base.py and run_pahfit script.
- Residuals panel added
- Errorbars with uncertainties added
- Tickmarks updated
M101_Nucleus_irs

### DIFF
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -310,8 +310,8 @@ class PAHFITBase:
         ax = axs[1]
         ax.minorticks_on()
         ax.tick_params(which='both', right='on', top='on', direction='in')
-        ax.plot(x, res, color='k')
-        ax.axhline(0, linestyle='--', color='gray')
+        ax.axhline(0, linestyle='--', color='gray', zorder=0)
+        ax.plot(x, res, color='k', zorder=1)
 
         ax.set_ylim(-2 * std, 2 * std)
         ax.set_xlabel(r"$\lambda$ [$\mu m$]")
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -262,7 +262,7 @@ class PAHFITBase:
         # spectrum and best fit model
         ax = axs[0]
         ax.minorticks_on()
-        ax.tick_params(which='both', top='on', direction='in')
+        ax.tick_params(which='both', right='on', top='on', direction='in')
         ax.plot(x, model(x) / x, "g-")
         ax.errorbar(x, y / x, yerr=yerr / x, 
                     fmt='o', markeredgecolor='k', markerfacecolor='none', 
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -313,8 +313,6 @@ class PAHFITBase:
         ax.set_yscale("linear")
         ax.set_xscale("log")
 
-        fig.subplots_adjust(hspace=0)
-
     def save(self, obs_fit, filename, outform):
         """
         Save the model parameters to a user defined file format.
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -311,7 +311,7 @@ class PAHFITBase:
         ax.minorticks_on()
         ax.tick_params(which='both', right='on', top='on', direction='in')
         ax.plot(x, res, color='k')
-        ax.hlines(0, linestyle='--', color='darkgray')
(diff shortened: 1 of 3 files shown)


karllark (commented):
Looks great.

I do wonder if the having the residual plot being a bit smaller (maybe height ratios of 3 to 1?) might be better.  

Also setting the default y-scale for the residuals to something other than min/max would be really nice.  Right now the very large noise at the edge of the wavelength range makes it hard to see the majority of the residuals.  Maybe compute the standard deviation of the residuals and then base the y-axis range to be +/- n*stddev where n would be 2 or 3?

alexmaragko:
> Looks great.
> 
> I do wonder if the having the residual plot being a bit smaller (maybe height ratios of 3 to 1?) might be better.
> 
> Also setting the default y-scale for the residuals to something other than min/max would be really nice. Right now the very large noise at the edge of the wavelength range makes it hard to see the majority of the residuals. Maybe compute the standard deviation of the residuals and then base the y-axis range to be +/- n*stddev where n would be 2 or 3?

That's a good suggestion. I'll work on those first chance.

alexmaragko:
M101_Nucleus_irs
- Height ratio set to 3:1
- ylim +/- 2*std of the residuals

ThomasSYLai:
@alexmaragko Thanks for adding this function! This can be very helpful. I have a small suggestion here. Perhaps adding a horizontal line with light color at residual=0 will help us better spot places with larger discrepancies.

alexmaragko:
Sure thing! By the way any idea why the codestyle test is failing? It states "pahfit --count --max-line-length=100" but I'm not sure where this is happening.

karllark:
> By the way any idea why the codestyle test is failing? It states "pahfit --count --max-line-length=100" but I'm not sure where this is happening.

The four lines above that state what lines have codestyle errors.

```
pahfit/base.py:271:28: E261 at least two spaces before inline comment
pahfit/base.py:313:29: E226 missing whitespace around arithmetic operator
pahfit/base.py:320:23: E226 missing whitespace around arithmetic operator
pahfit/base.py:320:30: E226 missing whitespace around arithmetic operator
```

karllark:
You can lookup the E??? values to learn the details.

alexmaragko:
M101_Nucleus_irs

pahfit/base.py
```diff
@@ -287,8 +300,24 @@ def plot(self, ax, x, y, model):
...
+        # residuals
+        res = (y - model(x)) / x
+        std = np.std(res)
+        ax = axs[1]
+        ax.minorticks_on()
+        ax.tick_params(which='both', right='on', top='on', direction='in')
+        ax.axhline(0, linestyle='--', color='gray', zorder=0)
+        ax.plot(x, res, color='k', zorder=1)
+
+        ax.set_ylim(-2 * std, 2 * std)
```
karllark, line 316: It would be great if you could put in the scale factor (here 2) as a keyword input to this function (e.g., `scalefac_resid=2` in the command line) so that it can be changed by users.

pahfit/scripts/run_pahfit.py
```diff
@@ -137,18 +137,19 @@ def main():
...
 
-    pmodel.plot(ax, obs_x, obs_y, obs_fit)
-
-    ax.set_yscale("linear")
-    ax.set_xscale("log")
+    pmodel.plot(axs, obs_x, obs_y, obs_unc.value, obs_fit)
 
     # use the whitespace better
-    fig.tight_layout()
+    fig.subplots_adjust(hspace=0)
```
karllark, line 152: Why the change from tight_layout to subplots_adjust?  Not a problem, I'm curious and want to learn.

karllark (changes requested):
Looks awesome.  Very minor change request.

pahfit/base.py
alexmaragko: Sure! I'll try to do it sometime today or tomorrow and also update plot_pahfit.py script

pahfit/scripts/run_pahfit.py
alexmaragko: I had fig.subplots_adjust(hspace=0) above the tight_layout line without removing the tiight_layout() but then tight_layout overrode hspace=0 and made the panels with an offset.

karllark (approved):
Looks awesome!  Thanks.

PR #117, opened 2021-04-23
Fix right axis

alexmaragko:
This PR fixes a double entry of the right side attenuation axis introduced after PR #115 merge. This was caused because I included the right side attenuation axis code (introduced in PR #114) independently in PR #115, because PR #115 was created from a branch before the merge of PR #114. 

Probably this is an example where things can be tricky with multiple pending PRs.

### DIFF
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -258,7 +258,7 @@ class PAHFITBase:
             observed spectrum uncertainties
         model : PAHFITBase model
             model giving all the components and parameters
-        scalefac_resid : integer
+        scalefac_resid : float
             Factor multiplying the standard deviation of the residuals to adjust plot limits
         """
         # spectrum and best fit model
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -273,16 +273,11 @@ class PAHFITBase:
         ax_att = ax.twinx()  # axis for plotting the extinction curve
         ax_att.tick_params(direction='in')
 
-        ax_att = ax.twinx()  # axis for plotting the extinction curve
-
         # get the extinction model (probably a better way to do this)
         for cmodel in model:
             if isinstance(cmodel, S07_attenuation):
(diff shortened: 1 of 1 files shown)


karllark:
Thanks for catching this.  Is is ready to merge?

alexmaragko:
Yes it is ready to merge. I would be also happy to know what is the proper way to deal with such cases. Should I have made the PR without including the PR #114 code? If so the right side axis in my PR would not be present. Perhaps there is a way to update the branch you are working on with the upstream without loosing your work (?)

karllark:
Thanks.  Will merge.

Yes, there is a way to handle this case.  You can always rebase a branch to the latest master branch.  You can read up on rebasing, it is a pretty standard technique.  I do this with these commands:

```
git checkout master
git fetch upstream
git merge upstream/master
git checkout <branch_name>
git rebase master
```
Note that conflicts can happen that you need to manually fix.  The git rebase command tells you how to do this and how to resume the rebase.

After such rebasing, you'll need to "force" your push to github as you are rewriting the commit history of your branch.  The `git push` command will fail and tell you how to do this.  (just adding a `--force` to the command if my memory serves).

karllark (approved):
Looks great.

alexmaragko:
> Thanks. Will merge.
> 
> Yes, there is a way to handle this case. You can always rebase a branch to the latest master branch. You can read up on rebasing, it is a pretty standard technique. I do this with these commands:
> 
> ```
> git checkout master
> git fetch upstream
> git merge upstream/master
> git checkout <branch_name>
> git rebase master
> ```
> 
> Note that conflicts can happen that you need to manually fix. The git rebase command tells you how to do this and how to resume the rebase.
> 
> After such rebasing, you'll need to "force" your push to github as you are rewriting the commit history of your branch. The `git push` command will fail and tell you how to do this. (just adding a `--force` to the command if my memory serves).

Good to know, thank you!

PR #123, opened 2021-05-26
fixed typo in article ref in component_models.py

els1:
Fixed typo:
Kemper, Vriend, & Tielens instead of Kemper, Vreind, & Tielens.


karllark (approved):
Looks great!

PR #133, opened 2021-06-14
Add M101_Nucleus_irs_IDL-PAHFIT-result.xdr

els1:
Result of IDL PAHFIT (using defaults) for 'M101_Nucleus_irs.ipac' in IDL's xdr format. File is added to data dir.


karllark:
I can see the value of having this file in the repository.  But, how does one read xdr format in python?  Is this possible?  Might want to add a small README for this file in the same directory so that future us remembers the details.

PR #136, opened 2021-06-14
replace estimate_start with no_starting_estimate

alexmaragko:
Replacing `estimate_start` with `no_starting_estimate` argument, to bypass the start point estimation if needed when fitting.


karllark (approved):
Looks great!

PR #138, opened 2021-06-21
Add doc on version description and PAHFIT v2.1 summary

ThomasSYLai:
Documentation added:

- Version description
- PAHFIT v2.1 summary on newly added features


docs/PAHFIT_v2.1.rst
```diff
@@ -0,0 +1,74 @@
...
+"""""""""""""""""""""""""""""""""""""""""""""""
+To run the fit and output the plot, simply type:
+
+.. code-block:: console
+
+  $ run_pahfit 1C_snr_gt20.ecsv scipack_ExGal_AkariIRC+SpitzerIRSSLLL.ipac --estimate_start
+
+.. code-block:: console
+
+  $ plot_pahfit 1C_snr_gt20.ecsv 1C_snr_gt20_output.ipac
```
karllark, line 72: It would be great to have a plot of the results to illustrate the awesomeness of adding AKARI.  Is this something you could add to this PR (like we already do in plot_result.rst)?

docs/version_description.rst
```diff
@@ -0,0 +1,35 @@
...
+
+v1.0 - v1.4
+------------
+
+The original IDL version of PAHFIT can be found in: `http://tir.astro.utoledo.edu/jdsmith/research/pahfit.php <http://tir.astro.utoledo.edu/jdsmith/research/pahfit.php>`_. For more details and background of ``PAHFIT``, see the `background <https://pahfit.readthedocs.io/en/latest/background.html>`_ page.
+
+v2.0
+------------
+
+A direct translation of python PAHFIT from the IDL version based on the astropy.modeling framework. Also, the modular approach has been built-in for future contributions, e.g., science/instrument packs.
```
karllark, line 13: I don't think we should say it is a direct translation.  We have attempted to match the capabilities of the IDL version, but we didn't do this by directly translating the IDL code to python.  This is implied by the 2nd sentence as we have actually made significant changes to make the python version more flexible than the IDL version already.

docs/version_description.rst
```diff
@@ -0,0 +1,35 @@
...
+
+v2.3
+------------
+Adaptation of 2.1/2.2 to JWST simulated data
+
+v2.4
+------------
+Adaptation of 2.1/2.2 to JWST based on ERS data
+
+v2.5
```
karllark, line 33: No need to say what version each new version is based on.  I would instead just say what the goal is (e.g., v2.2 could be something like "Enhancement to support Infrared Space Observatory spectra").

karllark: And I wouldn't say for v2.4 (or v2.5) that is based only on JWST ERS data.  I expect we will be using more than just the ERS data for this version.

docs/version_description.rst
```diff
@@ -0,0 +1,35 @@
...
+------------
+Final ERS delivery
```
karllark, line 35: Just remove version 2.5.  I fully expect that the we will have more versions at this point.

karllark (changes requested):
Great!  Some changes requested.

ThomasSYLai:
This has been replaced by PR #170.

PR #141, opened 2021-07-06
Estimate feature strengths and uncertainties

alexmaragko:
This PR calculates the strengths and profile-weighted uncertainties of the dust features and lines, based on pahfit_main_feature_power.pro

### DIFF
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -365,7 +365,8 @@ class PAHFITBase:
         ax.xaxis.set_minor_formatter(mpl.ticker.ScalarFormatter())
         ax.xaxis.set_major_formatter(mpl.ticker.ScalarFormatter())
 
-    def save(self, obs_fit, x, yerr, filename, outform):
+    @staticmethod
+    def save(obs_fit, x, yerr, filename, outform):
         """
         Save the model parameters to a user defined file format.
 
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -4,6 +4,7 @@ from pahfit.component_models import BlackBody1D, S07_attenuation
 
 from astropy.table import Table, vstack
 from astropy.modeling.physical_models import Drude1D
+from astropy import constants as const
 
 from scipy import interpolate
 
@@ -364,7 +365,7 @@ class PAHFITBase:
         ax.xaxis.set_minor_formatter(mpl.ticker.ScalarFormatter())
         ax.xaxis.set_major_formatter(mpl.ticker.ScalarFormatter())
 
-    def save(self, obs_fit, filename, outform):
+    def save(self, obs_fit, x, yerr, filename, outform):
         """
         Save the model parameters to a user defined file format.
 
@@ -421,6 +422,8 @@ class PAHFITBase:
                 "fwhm_min",
                 "fwhm_max",
                 "fwhm_fixed",
+                "strength",
+                "strength_unc"
             ),
             dtype=(
                 "U25",
@@ -437,6 +440,8 @@ class PAHFITBase:
                 "float64",
                 "float64",
                 "bool",
+                "float64",
+                "float64"
             ),
         )
         att_table = Table(
@@ -463,6 +468,19 @@ class PAHFITBase:
                     ]
                 )
             elif comp_type == "Drude1D":
+
+                # Calculate feature strength.
+                # TODO: Use input units to determine conversion and feature strength units.
+                strength = (np.pi * const.c.to('micron/s').value / 2) * (component.amplitude.value * component.fwhm.value / component.x_0.value**2) * 1e-26
+                # Estimate profile-weighted uncertainty.
+                if strength != 0.:
+                    nu = const.c.to('micron/s').value / x.value
+                    dnu = np.diff(nu, prepend=nu[1])
+                    uwv = np.where(component(x.value) / np.nanmax(component(x.value)) > 0.1)[0]
+                    strength_unc = np.sqrt(np.sum(np.take(yerr, uwv)**2 * np.take(dnu, uwv)**2)) * 1e-26
+                else:
+                    strength_unc = 0.
+
                 line_table.add_row(
                     [
                         component.name,
@@ -479,9 +497,25 @@ class PAHFITBase:
                         component.fwhm.bounds[0],
                         component.fwhm.bounds[1],
                         component.fwhm.fixed,
+                        strength,
+                        strength_unc,
                     ]
                 )
             elif comp_type == "Gaussian1D":
+
+                # Calculate line strength.
+                # TODO: Use input units to determine conversion and line strength units.
+                fwhm = 2 * component.stddev.value * np.sqrt(2 * np.log(2))
+                strength = ((np.sqrt(np.pi / np.log(2)) / np.pi) * const.c.to('micron/s').value / 2) * (component.amplitude.value * fwhm / component.mean.value**2) * 1e-26
+                # Estimate profile-weighted uncertainty.
+                if strength != 0.:
+                    nu = const.c.to('micron/s').value / x.value
+                    dnu = np.diff(nu, prepend=nu[1])
+                    uwv = np.where(component(x.value) / np.nanmax(component(x.value)) > 0.1)[0]
+                    strength_unc = np.sqrt(np.sum(np.take(yerr, uwv)**2 * np.take(dnu, uwv)**2)) * 1e-26
+                else:
+                    strength_unc = 0.
+
                 line_table.add_row(
                     [
                         component.name,
@@ -498,6 +532,8 @@ class PAHFITBase:
                         2.355 * component.stddev.bounds[0],
                         2.355 * component.stddev.bounds[1],
                         component.stddev.fixed,
+                        strength,
+                        strength_unc,
                     ]
                 )
             elif comp_type == "S07_attenuation":
--- a/pahfit/scripts/run_pahfit.py
+++ b/pahfit/scripts/run_pahfit.py
@@ -91,7 +91,7 @@ def main():
 
     # save fit results to file
     outputname = args.spectrumfile.split(".")[0]
-    pmodel.save(obsfit, outputname, args.saveoutput)
+    pmodel.save(obsfit, obsdata["x"], obsdata["unc"], outputname, args.saveoutput)
 
     # plot result
     fontsize = 18


alexmaragko:
I tried to enable astropy fitting uncertainties estimation `fit = LevMarLSQFitter(calc_uncertainties=True)` in helpers.py to calculate eventually feature strength uncertainties from the parameters, but it returns no result (or error) when printing for example component.std. It's strange because I can get uncertainties when trying a simple gaussian fitting for example in ipython. 

The astropy PR which enabled uncertainties estimation (https://github.com/astropy/astropy/pull/10552) mentions "This does not work with units yet" but I don't think this is the issue in our case as we build the model without using units.

Any thoughts?

karllark:
@els1 and @jdtsmith : ping.  Please review this and the other two related PRs hopefully soonish.

pahfit/base.py
```diff
@@ -463,6 +469,19 @@ def save(self, obs_fit, filename, outform):
                     ]
                 )
             elif comp_type == "Drude1D":
+
+                # Calculate feature strength.
+                # TODO: Use input units to determine conversion and feature strength units.
```
els1, line 474: Note TODO in case of future changes to units to which input spectrum is converted.

pahfit/base.py
```diff
@@ -479,9 +498,25 @@ def save(self, obs_fit, filename, outform):
...
                         component.fwhm.bounds[1],
                         component.fwhm.fixed,
+                        strength,
+                        strength_unc,
                     ]
                 )
             elif comp_type == "Gaussian1D":
+
+                # Calculate line strength.
+                # TODO: Use input units to determine conversion and line strength units.
```
els1, line 508: Similar TODO note as above.

pahfit/base.py
jdtsmith: We should probably abstract away all unit conversion (input -> model -> output) from the main fitting code itself.  I think we decided to standardize on f_nu units, and their surface brightness equivalents (per sr), but otherwise any combination should be allowed.  I'll mention in an issue.

pahfit/base.py
```diff
@@ -463,6 +469,19 @@ def save(self, obs_fit, filename, outform):
...
                 )
             elif comp_type == "Drude1D":
+
+                # Calculate feature strength.
+                # TODO: Use input units to determine conversion and feature strength units.
+                strength = (np.pi * const.c.to('micron/s').value / 2) * (component.amplitude.value * component.fwhm.value / component.x_0.value**2) * 1e-26
+                # Estimate profile-weighted uncertainty.
+                if strength != 0.:
+                    nu = const.c.to('micron/s').value / x.value
+                    dnu = np.diff(nu, prepend=nu[1])
```
ThomasSYLai, line 479: Shouldn't this prepend nu[0]?

ThomasSYLai (changes requested):
Just need a simple check on whether nu is written correctly.

pahfit/base.py
alexmaragko: It shouldn't make a difference if the element you are prepending is the 0th or 1st.

pahfit/base.py
```diff
@@ -479,9 +498,25 @@ def save(self, obs_fit, filename, outform):
...
+                        strength,
+                        strength_unc,
                     ]
                 )
             elif comp_type == "Gaussian1D":
+
+                # Calculate line strength.
+                # TODO: Use input units to determine conversion and line strength units.
+                fwhm = 2 * component.stddev.value * np.sqrt(2 * np.log(2))
+                strength = ((np.sqrt(np.pi / np.log(2)) / np.pi) * const.c.to('micron/s').value / 2) * (component.amplitude.value * fwhm / component.mean.value**2) * 1e-26
```
els1, line 510: Shouldn't it just be sqrt of  pi / {ln(2)*2}? If so, one should not divide again by pi.

pahfit/base.py
```diff
@@ -463,6 +469,19 @@ def save(self, obs_fit, filename, outform):
...
+
+                # Calculate feature strength.
+                # TODO: Use input units to determine conversion and feature strength units.
+                strength = (np.pi * const.c.to('micron/s').value / 2) * (component.amplitude.value * component.fwhm.value / component.x_0.value**2) * 1e-26
+                # Estimate profile-weighted uncertainty.
+                if strength != 0.:
+                    nu = const.c.to('micron/s').value / x.value
+                    dnu = np.diff(nu, prepend=nu[1])
+                    uwv = np.where(component(x.value) / np.nanmax(component(x.value)) > 0.1)[0]
+                    strength_unc = np.sqrt(np.sum(np.take(yerr, uwv)**2 * np.take(dnu, uwv)**2)) * 1e-26
```
els1, line 481: This is similar to estimate 1 in the IDL code: pahfit_main_feature_power.pro which calculates the strength of the main PAH features (combining multiple Drude profiles where necessary). This is different from the uncertainty calculated for each Drude profile in pahfit.pro which is given as the standard output for IDL pahfit. 
Should we be consistent with IDL pahfit and use the latter instead here?
Also, a bit more documentation about the method used would be helpful (e.g. copied from PAHFIT main feature power " profile-weighted quadrature summed sigma over the range of interest, W/m^2").

pahfit/base.py
```diff
@@ -479,9 +498,25 @@ def save(self, obs_fit, filename, outform):
...
+                # Calculate line strength.
+                # TODO: Use input units to determine conversion and line strength units.
+                fwhm = 2 * component.stddev.value * np.sqrt(2 * np.log(2))
+                strength = ((np.sqrt(np.pi / np.log(2)) / np.pi) * const.c.to('micron/s').value / 2) * (component.amplitude.value * fwhm / component.mean.value**2) * 1e-26
+                # Estimate profile-weighted uncertainty.
+                if strength != 0.:
+                    nu = const.c.to('micron/s').value / x.value
+                    dnu = np.diff(nu, prepend=nu[1])
+                    uwv = np.where(component(x.value) / np.nanmax(component(x.value)) > 0.1)[0]
+                    strength_unc = np.sqrt(np.sum(np.take(yerr, uwv)**2 * np.take(dnu, uwv)**2)) * 1e-26
```
els1, line 516: same comment as above about the method (vs IDL pahfit).

els1 (changes requested):
See comments above.

pahfit/base.py
els1: Do I miss something? If you prepend the 0th, the first element in np.diff will be 0 while if you prepend the 1st, it will be (1st-0th)?

pahfit/base.py
jdtsmith: I do suggest we leave all "residuals from the model" uncertainty estimates out of the main core code.  That was included as a bit of a hack in IDL PAHFIT, several years after it was first written, when I noticed what a poor job pipeline uncertainties (and hence PAHFIT-derived feature uncertainties) were doing at moderate to high SNR.  Since this will depend on the quality of the model, whether input spectra are sensible for PAHFIT'ing, instrumental details, etc., I think we should get pipeline uncertainties working cleanly with the full covariance matrix, and then perhaps add another convenience routine, maybe something like `alt_uncertainty`, for people who'd like alternative uncertainty estimates after the fitting.

pahfit/base.py
alexmaragko: Yes, that's the idea.

pahfit/base.py
alexmaragko: > This is similar to estimate 1 in the IDL code: pahfit_main_feature_power.pro which calculates the strength of the main PAH features (combining multiple Drude profiles where necessary). This is different from the uncertainty calculated for each Drude profile in pahfit.pro which is given as the standard output for IDL pahfit.
> Should we be consistent with IDL pahfit and use the latter instead here?
> Also, a bit more documentation about the method used would be helpful (e.g. copied from PAHFIT main feature power " profile-weighted quadrature summed sigma over the range of interest, W/m^2").

As we have extensively discussed yesterday, these are the only type of uncertainties that we can calculate at the moment given that the uncertainties from the model are not returned, and as JD mentioned above, this can become and alternative/optional uncertainty estimation in the future. So think of this as a temporary solution as well as an additional feature. This uncertainty calculation can be applied to the individual PAH components as well as the combined ones.

pahfit/base.py
alexmaragko: I think you are right. If the others agree, I can delete and push the change.

pahfit/base.py
els1: Thanks. I can see that. As JD mentioned, should it be included in the main code for now or do we already split it off in another one?

pahfit/base.py
karllark: No idea myself.  @jdtsmith?  @ThomasSYLai?

pahfit/base.py
karllark: Can it be split off now, but then used by default?

karllark: Until we have a better solution?

karllark:
I feel that the solution to getting uncertainties returned from the fitting (via the covariance matrix) is now beyond the scope of this PR.  If the scipy fitter returns a covariance matrix, then uncertainties are available, hence the code in this PR is doing the right thing.  We should move forward with getting this PR merged and fix the uncertainties later.

pahfit/base.py
alexmaragko: > Can it be split off now, but then used by default?

Do you mean have the feature strength calculation as a separate module/file imported in base.py?

karllark: Probably better to have it imported and called in run_pahfit.py.

alexmaragko: I could set it up as a separate module, and then run_pahfit can pass siimply an additional argument (e.g. strength_calc=True) to the save function, and the save function would then import and use the new module. Does it make sense?

karllark: Yes.  By default, it should run the strength calc.

pahfit/base.py
jdtsmith: Yeah we want to leave "alternative estimates" fully out, even as a temporary work-around.  Of course this puts the burden on us to solve the "missing LM parameter uncertainties" problem, but that is absolutely crucial, so keeping that burden intact is fully appropriate !

pahfit/base.py
jdtsmith: This should quickly fall out of comparisons with IDL PAHFIT main_feature_strength results.  Thomas and I have some new "synthetic" uncertainty-free PAH models we are working on, which should provide a nice comparison point (including Akari wavelengths).

alexmaragko: @jdtsmith Regarding what is currently in the IDL version, which is the aim of this PR, do you agree that it shouldn't divide again by pi? I'm only trying to move this PR forward as @karllark also stated. If there is a different approach, we should probably discuss it in an issue, or make a separate PR.

pahfit/base.py
alexmaragko: No problem. I can remove the profile weighted uncetainties out, and re-introduce them if needed in a future PR.

PR #142, opened 2021-07-27
Combined feature strengths

alexmaragko:
This PR is based on PR #141 and expands on it, by adding the combined dust feature strengths calculation.
The strengths are appended in the output file.
Currently pending are the uncertainties and eqw calculations which will be added soon.


alexmaragko:
For the uncertainties there is an issue recovering the astropy fit uncertainties and covariances. See also https://github.com/PAHFIT/pahfit/pull/141#issuecomment-878868364. While this works on a single component model, it does not return std or cov for the pahfit model.

pahfit/base.py
```diff
@@ -512,14 +546,71 @@ def save(self, obs_fit, filename, outform):
...
+        -------
+        cftable : data table
+            Astropy table containing combined dust feature strengths.
+        """
+        # Define combined strength dictionary.
+        cfdic = {'PAH_62': {'range': [6.2, 6.3]},
+                 'PAH_77_C': {'range': [7.3, 7.9]},
+                 'PAH_83': {'range': [8.3, 8.4]},
+                 'PAH_86': {'range': [8.6, 8.7]},
+                 'PAH_112': {'range': [11.2, 11.4]},
```
els1, line 580: To be consistent with IDL PAHFIT and distinctive of the 11.2 Drude profile, please use label 'PAH_112_C'

els1 (changes requested):

els1:
Following the discussion in PR #141, this part of code was not included in IDL's pahfit.pro (the main code) but in a separate program. Should we keep these separate in python or not? @jdtsmith your thoughts?

jdtsmith:
I think this should be a separate code that operates on the fit result object/table.  This way you can re-run it with tables saved to disk, for example, without re-fitting.  There are probably a few other pieces of functionality that would be better broken off to other modules for flexibility.  Definitely worth some thought...

karllark:
I like the idea of having it as a separate function.  I would advocate for this function to be run by default and the results added to the output table.

alexmaragko:
> I like the idea of having it as a separate function. I would advocate for this function to be run by default and the results added to the output table.

Currently the combined feature strengths calculation is implemented as a static method (featcombine) and is run by default with the results added to the output table.

pahfit/base.py
ThomasSYLai, line 580: Note that in PAHFIT the combined PAH feature at 11 µm is dubbed "11.3 complex". For consistency, might be better to name it "PAH_113_C". If we decide to stick with "PAH_112_C", will need to mention it somewhere to avoid confusion.

PR #143, opened 2021-07-29
EQW calculation

alexmaragko:
This PR is based on PRs #141 and #142 and further adds the equivalent width calculation of the separate components, as well as the eqws of the combined dust features.


alexmaragko:
The strength uncertainties is the only thing pending now as discussed in https://github.com/PAHFIT/pahfit/pull/141#issuecomment-878868364 and https://github.com/PAHFIT/pahfit/pull/142#issuecomment-887441890.

alexmaragko:
I might have an idea on why the fit uncertainties are not returned. It is probably because the combined pahfit model is not a pure astropy compound model, because we are using our own BlackBody1D and S07_attenuation models.

karllark:
> I might have an idea on why the fit uncertainties are not returned. It is probably because the combined pahfit model is not a pure astropy compound model, because we are using our own BlackBody1D and S07_attenuation models.

That sounds reasonable.  Is there a way for us to get the uncertainties for the covariance matrix that is returned (I think) from with the fit?

alexmaragko:
What if we turned BlackBody1D and S07_attenuation into an astropy [custom_model](https://docs.astropy.org/en/stable/api/astropy.modeling.custom_model.html#astropy.modeling.custom_model)?

karllark:
> What if we turned BlackBody1D and S07_attenuation into an astropy [custom_model](https://docs.astropy.org/en/stable/api/astropy.modeling.custom_model.html#astropy.modeling.custom_model)?

Maybe.  I don't know.  Is there something we could add to our custom models that would allow the covariance matrix to be computed?  Maybe we are just missing some property we could add that would allow things to work.  If custom_model works, then this might be what is happening.

alexmaragko:
https://github.com/alexmaragko/test_files/blob/main/test_astropy_compound_models_v01.py

This is a test script creating different compound models from "pure" Astropy models or combination of "pure" + custom models.

The parameter std seems to be returned only for the "pure' compound model.

pahfit/base.py
```diff
@@ -512,14 +570,139 @@ def save(self, obs_fit, filename, outform):
...
+            gauss = Gaussian1D(amplitude=amp,
+                               mean=x_0,
+                               stddev=fwhm_stddev)
+            lnu = gauss(lam)
+
+        # Set default broad limit value.
+        bl = 0.05
+
+        # Calculate EQW.
+        if fwhm > bl:
```
els1, line 697: In IDL PAHFIT, the condition is fractional FWHM > bl .... should it be the same here?

pahfit/base.py
```diff
@@ -512,14 +570,139 @@ def save(self, obs_fit, filename, outform):
...
+                            fwhm=fwhm)
+            lnu = drude(lam)
+
+        elif comp_type == 'Gaussian1D':
+            gauss = Gaussian1D(amplitude=amp,
+                               mean=x_0,
+                               stddev=fwhm_stddev)
+            lnu = gauss(lam)
+
+        # Set default broad limit value.
```
els1, line 693: I would copy the documentation/explanation given for this and how it is used in IDL PAHFIT (copied above) so we know 1yr from now what we did and why without having to dig in the IDL code.

els1 (changes requested):

pahfit/base.py
```diff
@@ -512,14 +570,139 @@ def save(self, obs_fit, filename, outform):
...
+        ----------
+        comp_type : string
+            type of emission component (Drude1D/Gaussian)
+        x_0 : float
+            central wavelength of the feature.
+        amp : float
+            central intensity of the feature.
+        fwhm_stddev : float
+            fwhm or stddev of the feature depending on comp_type.
+
```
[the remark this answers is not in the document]
jdtsmith: Honestly I never trusted the EQW measured by main_feature_power.  I personally don't like EQW measures and don't use them (typically), but I also think this is squarely in the domain of "post-processing" that shouldn't be in the core fitting package.  A big picture way to think about it: we want to avoid making any scientific choices for end users, other than those expressed in our science packs.

karllark:
I think I'll open an issue to discuss the uncertainties.  Just so there is a place to discuss this outside of this and the other PRs.  I'm thinking that the solution may happen outside of these PRs.  But I could be wrong.  Will link the issue to all these PRs so we maintain the connection.

pahfit/base.py
alexmaragko: It depends if `st.fwhm` in line 598 of pahfit.pro is the standard fwhm or the fractional. If it is the fractional then yes I should change this here.

PR #144, opened 2021-07-29
Not flux

BethanyRS:
In my files comparing the IDL and Python fit, what I have been calling the 'flux' should be called the 'integrated intensity'.  My png and rst files should now be updated accordingly.


BethanyRS:
I ended up switching the x and y pixel values to better match the article in the rst.  All values from my previous PR are still the same.  The only differences are my strings containing the pixel names.

karllark (approved):
Looks great.  Thanks.

PR #145, opened 2021-08-13
Fix error when tau=0 and Fixed

### LINKED ISSUE #112: Issue with run_pahfit when S07_att's amp = 0.0 and amp_fixed = True
Setting amp = 0.0 and amp_fixed = True for S07_att (the last row in scipack_ExGal_SpitzerIRSSLLL.ipac) results in the output

"The cosine of the angle between func(x) and any column of the
  Jacobian is at most 0.000000 in absolute value"

karllark:
Tau=0 and fixed in the packfile causes the fitting to not happen.  This is due to a bug in the S07 model that returned a vector of zeros instead of ones when tau=0.  This PR changes the behavior to return ones.

Closes #112.


jdtsmith (approved):
Looks good.

PR #151, opened 2021-10-19
Estimate feature strengths

alexmaragko:
This PR should replace #141. 

Differences with #141:
- New separate module `feature_strengths.py` now called to calculate PAH feature and emission line strengths.
- Profile-weighted uncertainties removed, but output table uncertainty entry kept (empty) as placeholder for future implementation.
- New argument in `run_pahfit` (`--strength_calc`) which will enable feature strength calculation when provided.

### DIFF
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -367,7 +367,7 @@ class PAHFITBase:
         ax.xaxis.set_major_formatter(mpl.ticker.ScalarFormatter())
 
     @staticmethod
-    def save(obs_fit, filename, outform, strength_calc=True):
+    def save(obs_fit, filename, outform):
         """
         Save the model parameters to a user defined file format.
 
@@ -380,9 +380,6 @@ class PAHFITBase:
             Currently using the input data file name.
         outform : string
             Sets the output file format (ascii, fits, csv, etc.).
-        strength_calc : bool
-            Imports the feature_strength module
-            to calculate PAH feature and emission line strenghts.
         """
         # setup the tables for the different components
         bb_table = Table(
@@ -474,12 +471,9 @@ class PAHFITBase:
                 )
             elif comp_type == "Drude1D":
 
-                if strength_calc:
-                    strength = pah_feature_strength(component.amplitude.value,
-                                                    component.fwhm.value,
-                                                    component.x_0.value)
-                else:
-                    strength = None
+                strength = pah_feature_strength(component.amplitude.value,
+                                                component.fwhm.value,
+                                                component.x_0.value)
 
                 strength_unc = None
 
@@ -505,12 +499,9 @@ class PAHFITBase:
                 )
             elif comp_type == "Gaussian1D":
 
-                if strength_calc:
-                    strength = line_strength(component.amplitude.value,
-                                             component.mean.value,
-                                             component.stddev.value)
-                else:
-                    strength = None
+                strength = line_strength(component.amplitude.value,
+                                         component.mean.value,
+                                         component.stddev.value)
 
                 strength_unc = None
 
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -365,7 +365,8 @@ class PAHFITBase:
         ax.xaxis.set_minor_formatter(mpl.ticker.ScalarFormatter())
         ax.xaxis.set_major_formatter(mpl.ticker.ScalarFormatter())
 
-    def save(self, obs_fit, x, yerr, filename, outform):
+    @staticmethod
+    def save(obs_fit, x, yerr, filename, outform):
         """
         Save the model parameters to a user defined file format.
 
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -4,13 +4,15 @@ from pahfit.component_models import BlackBody1D, S07_attenuation
 
 from astropy.table import Table, vstack
 from astropy.modeling.physical_models import Drude1D
-from astropy import constants as const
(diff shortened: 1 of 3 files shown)


pahfit/feature_strengths.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Calculate emission line strengths.
+
+    Parameters
+    ----------
+    ampl : float
+        Gaussian1D amplitude
+    mean : float
+        Mean of the Gaussian
+    fwhm : float
```
els1, line 40: input is called 'stddev', not 'fwhm'

pahfit/feature_strengths.py
```diff
@@ -0,0 +1,54 @@
...
+def line_strength(ampl, mean, stddev):
+    """
+    Calculate emission line strengths.
+
+    Parameters
+    ----------
+    ampl : float
+        Gaussian1D amplitude
+    mean : float
+        Mean of the Gaussian
```
els1, line 39: mean in wavelength?

pahfit/base.py
```diff
@@ -11,6 +11,8 @@
 
 import matplotlib as mpl
 
+from pahfit.feature_strengths import feature_strength, line_strength
```
els1, line 14: very similar names feature_strengths and feature_strength

els1 (changes requested):
Only small comments. Overall looks good to me.

pahfit/feature_strengths.py
alexmaragko: Yes. I'm using the astropy description.

pahfit/base.py
alexmaragko: Changed the function to pah_feature_strengths.

els1 (approved):
Looks good.

pahfit/feature_strengths.py
jdtsmith: @els1 You are referring to x_0?

els1: yes

ThomasSYLai (approved):
Overall, this looks good to me.

pahfit/base.py
```diff
@@ -377,6 +380,9 @@ def save(self, obs_fit, filename, outform):
             Currently using the input data file name.
         outform : string
             Sets the output file format (ascii, fits, csv, etc.).
+        strength_calc : bool
+            Imports the feature_strength module
```
karllark, line 384: This import is done automatically.  I reword this description to say what this code is doing.  Like: "compute the integrated dust and gas line strengths."  That is what this code is doing, right?

pahfit/feature_strengths.py
```diff
@@ -0,0 +1,54 @@
+import numpy as np
+from astropy import constants as const
+
+
+def pah_feature_strength(ampl, fwhm, x_0):
+    """
+    Calculate PAH feature strengths.
```
karllark, line 7: Maybe a bit more description here?  Feature strengths could be the central intensity or the integrated area.

pahfit/scripts/run_pahfit.py
```diff
@@ -70,6 +70,11 @@ def initialize_parser():
         type=int,
         help="Maximum number of interations for the fitting",
     )
+    parser.add_argument(
+        "--strength_calc",
+        action="store_true",
+        help="Calculate PAH feature and emission line strenghts",
+    )
```
karllark, line 77: I would argue that computing the integrated feature strengths should be done by default, not only if the user requests.  This is not a computationally intensive calculation, so doing it by default will provide more information to users by default.  Hence this switch should be reversed to allow for the feature strengths to be turned off, not on.

pahfit/base.py
```diff
@@ -364,7 +366,8 @@ def plot(self, axs, x, y, yerr, model, scalefac_resid=2):
         ax.xaxis.set_minor_formatter(mpl.ticker.ScalarFormatter())
         ax.xaxis.set_major_formatter(mpl.ticker.ScalarFormatter())
 
-    def save(self, obs_fit, filename, outform):
+    @staticmethod
+    def save(obs_fit, filename, outform, strength_calc=True):
```
karllark, line 370: I like the default to be to do the strength calculation.

karllark (approved):
Looks good to me.  A few comments to improve the docs and usage.

pahfit/scripts/run_pahfit.py
alexmaragko: I totally agree with you. I only added an option to enable the calculation because it was mentioned at some point that we might not want to deliver measurements to the user. So since we are going to compute strengths by default, do we really need an option to override it? The output will include the feature strengths and the user can simply ignore them.

karllark: Good point.  I would remove the option to overwrite it.  The user can just ignore it if they aren't interested.

alexmaragko:
Feature strengths are now calculated by default. No argument required. Not sure why "link_test" canceled in the checks.

karllark:
Re-running the checks.  Sometimes they fail just because.  ;-)

karllark:
@alexmaragko: is this PR ready for merging in your view?  Have you responded to all the comments?

alexmaragko:
> @alexmaragko: is this PR ready for merging in your view? Have you responded to all the comments?

@karllark Yes I believe so.

karllark:
Thanks for the response.  There are multiple approvals so will merge.

Issue #19, opened 2018-10-19
PAHFIT Model Component Extraction


jdtsmith:
One of the most useful (and under-used) features of the original `PAHFIT` is the `pahfit_components` function, which looks like this:

```idl
function pahfit_components,lambda,decoded_params, $
                           DUST_CONTINUUM=dust_continuum, $
                           TOTAL_DUST_CONTINUUM=dc_tot,STARLIGHT=stars, $
                           TOTAL_CONTINUUM=tot_cont, $
                           DUST_FEATURES=dust_features, $
                           TOTAL_DUST_FEATURES=df_tot, $
                           LINES=lines,TOTAL_LINES=lines_tot, $
                           EXTINCTION_FAC=ext, $
                           EXTINCTION_CURVE=ext_curve, $
                           DISABLE_EXTINCTION=de,LAMBDA_REST=lam_rest
```

This code allows you to take a `decoded_params` structure, which is output from a `PAHFIT` run, and selectively pull pieces from it: the total continuum, all due features, each features separately, etc.  This is used to produce the canonical multi-colored plots we all known and love, but it's also essential for using model results more deeply than just examining the parameters output.  We need something similar.

karllark:
One option that is currently in the prototype base class is that each model (e.g., bb_cont, h2_model, etc.) is saved in the class as a separate compound model.  Will need code to update the parameters after the fitting is done, but otherwise you can just use the separate component models for this.  Will need to provide an example of this (docs!).

jdtsmith:
Interesting.  But for many components, you need to _blend_ them together to be useful.  For example: total dust continuum (vs. the per-T continua, which are not very physically meaningful).  

Two separate but related parts of original `PAHFIT` pertain to this:

```
;+
; NAME:
;
;   PAHFIT_FEATURE_STRENGTH_COMBINE
;
; PURPOSE:
;
;   Combine feature strengths as produced by PAHFIT, taking care of
;   covariance among the features being combined when computing the
;   uncertainty.
;+
; NAME:
;
;    PAHFIT_MAIN_FEATURE_POWER
;
; PURPOSE:
;
;    Compute the power of the main dust features from a decoded PAHFIT
;    structure.
```

These together combine PAH band strengths for PAH "Complexes".  Since they are all correlated, propagating errors is tricky.

karllark:
Good point.

jdtsmith:
[...]

PR #156, opened 2021-11-16
function to return model compounds

els1:
Added function to helpers.py to return the model compounds: total continuum, total dust continuum, total dust features, total H2 lines.

### DIFF
--- a/pahfit/helpers.py
+++ b/pahfit/helpers.py
@@ -243,3 +243,4 @@ def calculate_compounds(obsdata, pmodel):
     compounds["extinction_model"] = ext_model
 
     return compounds
+    
--- a/pahfit/helpers.py
+++ b/pahfit/helpers.py
@@ -145,10 +145,12 @@ def fit_spectrum(obsdata, pmodel, maxiter=1000, verbose=True):
     return obs_fit
 
 
-def get_compounds(obsdata, pmodel):
+def calculate_compounds(obsdata, pmodel):
     """
-    Determine model compounds for total continuum, stellar continuum, total dust continuum, combined dust features, 
-    combined atomic and H2 lines, combined H2 lines, combined atomic lines, and extinction model
+    Determine model compounds for total continuum, stellar continuum,
+    total dust continuum, combined dust features,
+    combined atomic and H2 lines, combined H2 lines,
+    combined atomic lines, and extinction model
 
     Parameters
     ----------
@@ -239,6 +241,5 @@ def get_compounds(obsdata, pmodel):
     compounds["h2_lines"] = h2_features_model(x)
     compounds["atomic_lines"] = atomic_features_model(x)
     compounds["extinction_model"] = ext_model
-
 
     return compounds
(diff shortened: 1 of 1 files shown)


els1:
Issue #19: IDL version gives (in addition to individual components)

TOTAL_DUST_CONTINUUM=dc_tot, TOTAL_CONTINUUM=tot_cont,                         TOTAL_DUST_FEATURES=df_tot, LINES=lines,TOTAL_LINES=lines_tot, $
EXTINCTION_CURVE=ext_curve

I didn't add:
total lines, only total H2 lines
extinction curve

Should this be added?

karllark:
The codacy failure is minor and is due to trailing white space on some lines.  This is a PEP8 thing.  It can be left or fixed if you want.  Click on the Details link and and then the view more details to see the specific lines that have this issue.

karllark:
I think you should add a compound that is the atomic emission lines at a minimum (currently this code returns the H2 only).  

I would return the attenuation curve as well.  Useful as in the future we will have a more complex attenuation curve (e.g., ices with the v2.1 AKARI version).

els1:
> The codacy failure is minor and is due to trailing white space on some lines. This is a PEP8 thing. It can be left or fixed if you want. Click on the Details link and and then the view more details to see the specific lines that have this issue.

Thanks. I think I removed them.

karllark (approved):
Looks good to me.

karllark:
One thought.  It would be good to have some documentation on this useful function.  Like a doc page that shows the output for our standard M101 nuclear fit.  This could be part of this PR.  But doesn't have to be.  If not, then an issue should be opened so that such docs are added at some point.

els1:
> One thought. It would be good to have some documentation on this useful function. Like a doc page that shows the output for our standard M101 nuclear fit. This could be part of this PR. But doesn't have to be. If not, then an issue should be opened so that such docs are added at some point.

I will work on including the docs in this PR.

pahfit/helpers.py
```diff
@@ -8,6 +8,11 @@
 
 from pahfit.base import PAHFITBase
 
+from pahfit.component_models import BlackBody1D, S07_attenuation
+from astropy.modeling.physical_models import Drude1D
+from astropy.modeling.functional_models import Gaussian1D
+
+
 __all__ = ["read_spectrum", "initialize_model", "fit_spectrum"]
```
ThomasSYLai, line 16: add "calculate_compounds" in __all__

ThomasSYLai (changes requested):
Everything looks good to me except adding "calculate_compounds" in __all__

pahfit/helpers.py
els1: corrected. thanks.

ThomasSYLai (approved):

els1:
I will add the docs in a new PR as there is an active PR on docs.

karllark:
So, is this PR ready for merging @els1?

els1:
@karllark yes it is.

PR #161, opened 2021-11-30
Implementation of Asymmetric Drude profiles

Ameek-Sidhu:
This PR implements asymmetric Drude profiles into PAHFIT.


karllark:
Given that the Drude is already asymmetric intrinsically, I would suggest we call the function "Drude_modified" or "ModDrude" or something like that.  In my 2021 paper where I introduced this variably asymmetric Drude I called it the "modified Drude" to avoid confusion about the intrinsic and "extra" asymmetry.

Ameek-Sidhu:
I have changed the name of the function to 'Drude_modified' and updated the pack file with asymmetry set to zero and a fixed parameter.

jdtsmith:
I think we should keep Drude1D as is, and use another `Drude1DAsymmetric` component for now.  Then we can decide, _ala carte_, if we'd like to turn on the asym version for particular features in future science packs.  This will require some significant training.  So for now, just implementing a `Drude1DAsymmetric`, but not yet putting it to use, makes the most sense to me.

jdtsmith:
Since we are likely to change the underlying model format soon, and we don't (yet) know if the asymmetric Drude will be worth the complexity it adds, I'll close this for now and we can revisit later.

Issue #160, opened 2021-11-22
update M101 output file


els1:
update docs: M101 output file in https://pahfit.readthedocs.io/en/latest/fit_spectrum.html
to reflect changes from PR #151

karllark:
Done in above PR.

Issue #157, opened 2021-11-16
Documentation save method in base.py


els1:
For input parameters, save method in base.py states:

filename : string
            String used to name the output file.
            Currently using the input data file name.

The requested filename does not include the file extension. Should this be mentioned explicitly?

karllark:
Yes.

Issue #137, opened 2021-06-14
Description output file incorrect for fit_spectrum


els1:
I think there is an incorrect description for the output file in fit_spectrum: 

The code states:
    Returns
    -------
    obsfit : PAHFITBase model
        PAHFIT model with best fit parameters

But if you do:
obsfit = fit_spectrum(obs, pmodel, maxiter=2000)
help(obsfit)

you get:
Help on CompoundModel in module astropy.modeling.core object:

class CompoundModel(Model)
 |  CompoundModel(op, left, right, name=None, inverse=None)
 |  
....

els1:
I think the same issue is in the docs for base.py: plot 

Parameters
 ----------
...
model : PAHFITBase model
            model giving all the components and parameters

but input is pmodel.model or the output of fit_spectrum.

karllark:
The PAHFITBase model is an astropy modeling CompoundModel.  This this is correct at least in a sense.  Maybe we need to make it clearer that PAHFITBase model is a CompoundModel?

els1:
Perhaps. 

What I referred to is that:
obsfit = fit_spectrum(obs, pmodel, maxiter=2000) 
and
pmodel = initialize_model(packs, obs, estimate_start=False)
both are listed in the docs as "PAHFITBase model"
while obsfit = pmodel.model
So I expected the docs to list them differently.

karllark:
Good point.

PR #162, opened 2021-12-07
Combined PAH feature strengths

alexmaragko:
This PR calculates the combined PAH feature strengths, and should replace #142.

Differences with #142:

- Combined PAH feature strength function is now inside `feature_strengths.py` module and called in `base.py`.

An EQW column is created in the output ipac table as placeholder for the next PR (that will replace #143).


pahfit/feature_strengths.py
```diff
@@ -53,3 +55,54 @@ def line_strength(ampl, mean, stddev):
...
+    -------
+    cftable : data table
+        Astropy table containing combined dust feature strengths.
+    """
+    # Define combined strength dictionary.
+    cfdic = {'PAH_62': {'range': [6.2, 6.3]},
+             'PAH_77_C': {'range': [7.3, 7.9]},
+             'PAH_83': {'range': [8.3, 8.4]},
+             'PAH_86': {'range': [8.6, 8.7]},
+             'PAH_112_C': {'range': [11.2, 11.4]},
```
ThomasSYLai, line 77: Copy from PR #142:
Note that in PAHFIT the combined PAH feature at 11 µm is dubbed "11.3 complex". For consistency, might be better to name it "PAH_113_C". If we decide to stick with "PAH_112_C", will need to mention it somewhere to avoid confusion.

ThomasSYLai (changes requested):

alexmaragko:
Last commit includes minor docstring updates, to address issues #157 and #137, and an update to the output table in fit_spectrum.rst to address #160.

Most of the test fail now because sphinx is not able to generate the rst file.

I have copied the ipac output table to the rst file, but it seems formatting issues appear. I'm not sure what is the quickest way to deal with this in order to avoid formating line by line. 

I have also tried the astropy ascii.rst format output which generates the table's body properly when pasted in the rst file but not the header.

alexmaragko:
Fixed the formatting issue by using the ascii.rst format output. Now the matrix test seems to be failing due to Python 3.6 setup with tox environment on macos.

pahfit/feature_strengths.py
karllark: Agree.

karllark (approved):
Looks good to me.  Modulo the naming of the 11.2/3 complex.

karllark:
Will work on a separate PR to update the text dependencies.

karllark:
Test failures fixed in PR #163.  Best solution would be to rebase this PR once that PR is merged.  That would allow the full tests to be run.  Ask and I can help through the 4-5 commands needed to do a rebase (should be clean and straightforward for this PR).

alexmaragko:
Rebased the PR and changed PAH_112_C to PAH_113_C.

PR #163, opened 2022-01-10
updating test versions

karllark:
python 3.6 no longer allowed.  python 3.7 old as well.  Moving tests to newer version of python, numpy, and astropy.  Based on current astropy tox.ini.

### DIFF
--- a/.github/workflows/tex-tests.yml
+++ b/.github/workflows/tex-tests.yml
@@ -27,7 +27,7 @@ jobs:
         # For example -- os: [ubuntu-latest, macos-latest, windows-latest]
         os: [ubuntu-latest, macos-latest, windows-latest]
 
-        # Test python 3.6, 3.7, and 3.8 by default.
+        # Test python 3.8, 3.9, and 3.10 by default.
         python-ver: [8, 9, 10]
 
         # Specify which tox environments to test in this list.
--- a/.github/workflows/tex-tests.yml
+++ b/.github/workflows/tex-tests.yml
@@ -28,7 +28,7 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
 
         # Test python 3.6, 3.7, and 3.8 by default.
-        python-ver: [6, 7, 8]
+        python-ver: [8, 9, 10]
 
         # Specify which tox environments to test in this list.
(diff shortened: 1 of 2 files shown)


.github/workflows/tex-tests.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
 
         # Test python 3.6, 3.7, and 3.8 by default.
```
els1, line 30: update the comment to test python 3.8, 3.9, 3.10 by default.

.github/workflows/tex-tests.yml
```diff
@@ -56,14 +56,14 @@ jobs:
     - name: Set up python for astropy, numpy dev test
       uses: actions/setup-python@v1
       with:
-        python-version: 3.8
+        python-version: '3.10'
```
els1, line 59: I assume the python version now needs be listed between ' ' (not done in previous version).

tox.ini
```diff
@@ -38,23 +38,25 @@ description =
...
     numpy118: numpy==1.18.*
+    numpy119: numpy==1.19.*
+    numpy120: numpy==1.20.*
+    numpy121: numpy==1.21.*
 
-    astropy30: astropy==3.0.*
     astropy40: astropy==4.0.*
-    astropylts: astropy==4.0.*
+    astropy50: astropy==5.0.*
+    astropylts: astropy==5.0.*
```
els1, line 59: on ln 47: astropalts: with the latest astropy LTS while here it is satrapy 5.0, ok?

els1 (changes requested):

.github/workflows/tex-tests.yml
karllark: Only if it ends in `0`.   Then 3.10 is interpreted as 3.1 causing an error.

.github/workflows/tex-tests.yml
karllark: Fixed.

tox.ini
karllark: The latest LTS version is 5.0.

karllark: Added this in the comment on line 47.

els1 (approved):

PR #167, opened 2022-02-17
Fix issue 150

### LINKED ISSUE #150: Can't fit data without 5.5 point
Lines 675 and 676 in base.py

( `else:  # if min(obs_x) > 5, use 5.5 um`
          `lam = 5.5`)

result in "ValueError: A value in x_new is below the interpolation range" when there isn't a point around 5.5 microns (e.g. if you're fitting a subset of data from 6-14 microns).

ThomasSYLai:
Fix #150. Now PAHFIT can fit spectrum with its shortest wavelength greater than 5.5 µm


drvdputt:
I have also run into this issue, and I think I made a temporary fix at some point. This does make me wonder: why does the 5.0 - 5.5 micron range need a different behavior than wav < 5.0 or wav > 5.5?

ThomasSYLai:
@drvdputt There is nothing special. The 5.5 µm was set in the original IDL PAHFIT. When developing the enhanced PAHFIT to fit akari+spitzer, I made it start at shorter wavelengths depending on the shortest available wavelength in akari (~2.6 µm). At some point we can probably just use min(obs_x) + 0.1  as the reference point. I assume the fitting of the continuum does not care about this reference wavelength that much as long as it's near the trough where the stellar component meets dust components.

karllark:
@BethanyRS : Please review this PR.  Thanks.

BethanyRS:
I'm still having a bit of trouble getting PAHFIT to run on a cut spectrum.  When I use a spectrum that ranges from 8-14 microns, I get the error "ValueError: max() arg is an empty sequence."  According to the error message, this occurs in line 77 of "component_models": 

`kvt_int_short = kvt_int_short_tmp * (kvt_int[0] / max(kvt_int_short_tmp))`  

Does this error have something to do with

`kvt_wav_short = in_x[in_x < min(kvt_wav)]`

(line 73 of "component_models")?  It looks like kvt_wav starts at 8 microns.

karllark:
@BethanyRS is this with Thomas' branch of pahfit?  This branch has not been merged yet, hence you would have to get his branch to test if this fix works.

BethanyRS:
Unless I'm misunderstanding how switching branches works, my comment should be for the branch of this PR

karllark:
> Unless I'm misunderstanding how switching branches works, my comment should be for the branch of this PR

Great thanks for confirming.  @ThomasSYLai can you comment on why @BethanyRS's test is not successful on this branch?

jdtsmith:
Looks like our parametric dust attenuation curve is brittle if you don't have shorter wavelength coverage.  This should be improved.

BethanyRS:
Just to touch base on this again, I'm still getting the same error when I cut my spectrum.


```
line 77, in kvt
    kvt_int_short = kvt_int_short_tmp * (kvt_int[0] / max(kvt_int_short_tmp))
ValueError: max() arg is an empty sequence

```

For reference, this time I took Orion_H2S1_ISO-SWS_merged in pahfit/pahfit/data/ISO_SWS and cut the data points so that they were between 8-14 microns. I tried to post the altered Orion_H2S1_ISO-SWS_merged here, but the file type wasn't supported. However, unless I'm misunderstanding something, I think that the same error will occur for any set of data that starts after 8 microns.

PR #175, opened 2022-03-30
Improve Plot density

jdtsmith:
This increases the density of the model over-plotting to sample Gaussians etc. better.  Looks much nicer (a trick I used in IDL PAHFIT).  It's configurable.


karllark (approved):
Nice improvement in the plot.  Great!

PR #176, opened 2022-03-30
Implement Modified Blackbody

### LINKED ISSUE #171: Blackbody continuum sum incorrectly implemented
Looking closely at the model implementation, it seems the sum of blackbody continua is incorrectly implemented.  Here is the PAHFIT model in simple form:

httptir astro utoledo edujdsmithresearchdownloadpah_template_final_apj pdf

Notice the B_nu/lam^2, which is a sum of emissivity=2 MBB's.  I believe the current model continuum sources are unmodified (which makes them "peakier" and thus a bit harder to define smooth continuum with). 

This brings up the need to validate against IDL PAHFIT not just in terms of the fit quality, but also the fit _parameters_ including the tau's in the above.  

I will take a crack at fixing this, probably by adding a "Modified" flag to the Blackbody1D model component and an associated column to the scipacks.  Or perhaps just make a `ModifiedBlackBody1D` derived model.  

Can someone take a look and confirm my suspicion here?

jdtsmith:
Formerly BlackBody1D was actually a modified blackbody.  With this fix, two separate components are created.  The sci-pack is updated accordingly.  Fixes #171.  

Improves the compositing of models to guard against missing components.  Formerly if no BB continua were set, the model setup would fail.


jdtsmith:
Looks like our "test" comparison to M101 nucleus hard-codes the outdated fit.  @karllark is that easy to update?

karllark:
Yes it is easy.  Look at `pahfit/test_fitting_spitzer.py`.  You need to do the fit with the new code and then print out the parameters as an array.  Use` print(repr(model.parameters))` to get something that is easy to cut and paste into the testing code.

karllark:
Ok.  Now I see it.  You removed the line that modified the blackbody and then put it in "ModifiedBlackBody".  Looks great.

karllark:
The PAHFITBase.write (and maybe read) functions need to be updated to handle the ModifiedBlackBody and output it.  This will address #180.

jdtsmith:
Fixed in 166d6dc.  Uses isinstance instead of comparing class string, so that component classes can be inherited.  We could for example make AsymmetricDrude1D a subclass of Drude1D.

karllark (approved):
Looks good.  Nice to have this fixed.

karllark:
Will merge and fix failing tests in a separate PR.

Issue #174, opened 2022-03-30
Gaussian1D line FWHM far too large


jdtsmith:
In the [nominal science pack](https://github.com/PAHFIT/pahfit/blob/7090b71e8ea183b3f64e028b7516f27d8b40f91c/pahfit/packs/scipack_ExGal_SpitzerIRSSLLL.ipac#L39), the lines have an incredibly wide 2x (below, above) latitude in FWHM.  This is far much, and does not match the IDL PAHFIT, where a 10% range was allowed.  This leads to absurdly narrow lines which the fitter "prefers" (since it goes perfectly "through" a couple points).  Example (notice ArII, H2 S(1), SiII):

2x FWHM constraints:
![IMG_3300](https://user-images.githubusercontent.com/93749/160918934-17b2b92d-10ea-4a05-bb14-c2a2bae7543b.png)

10% FWHM constraints
![IMG_2005](https://user-images.githubusercontent.com/93749/160918970-3861e541-8940-43ad-9a37-ede231992d19.png)
straints:

FWHM for unresolved lines should be the purview of the instrument pack, not the sci-pack.  We could add a "sigma_v" dispersion term that is allowed to be convolved with the Gaussian LSF specified in the sci-pack.  But that's for another day.  Right now I have re-hard-coded 10% constraints in the model reader, which works.  But a better fix will involve the sci/inst-pack separation.  Will submit workaround PR unless there are objections.

karllark:
You can  just update the scipack min/max values.  A bit tedious for sure.  But it is possible at least.

karllark:
Definitely a better fit with the tighter constraints.

karllark:
Fully agree.  Just thinking if you are looking for a "quick fix", you could just update the scipack file.

jdtsmith:
It's even quicker; see #177 :).

PR #177, opened 2022-03-30
Simple FWHM Fix

jdtsmith:
This is a simple fix for the overly generous 2x FWHM (see #174), which led to spurious fits.  This simply hard-codes the 10% for now.  In the near term, we plan to implement a new Instrument Pack/Science Pack with generator, making this code irrelevant.

### DIFF
--- a/pahfit/base.py
+++ b/pahfit/base.py
@@ -134,6 +134,9 @@ class PAHFITBase:
             # guess values and update starting point (if not set fixed) based on the input spectrum
             param_info = self.estimate_init(obs_x, obs_y, param_info)
 
+        if not param_info:
+            raise ValueError("No parameter information set.")
+        
         self.param_info = param_info
 
         bb_info = param_info[0]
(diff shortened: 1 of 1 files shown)


karllark:
Also needs the testing parameters update as this changes the results for M101.

jdtsmith:
I'm not as familiar with the test structure; can someone else take a look at that?

karllark:
For now, it might be best to accept failing tests and merge if we have good confidence that it is just needing to update the cached fit values.  There are now multiple PRs that have this test failure.  Then I can do a "quick" PR to fix the failing tests after the different merges are done. 

Or another solution is that I could provide the needed code snippet that would generate the lines that would need to be cut/pasted in to the test function.

Thoughts?

jdtsmith:
If we have a few ongoing changes in this category, perhaps “clean up after” is the right approach. I demoed my own branch at the workshop yesterday so probably good to get some of these in.

karllark (approved):
A simple fix.  Definitely need to fix this later "properly."

karllark:
Merging now.  Will fix failing tests in a separate PR.

PR #215, opened 2022-06-20
ISO-SWS Pack files

Ameek-Sidhu:
ISO-SWS pack files in ipac and yaml format. There are two ipac tables, one for the spectral resolution = 1600, and the other for the spectral resolution=500. The yaml file is the one created based on @jdtsmith's  PR #210. Please review this and let us (me and @BethanyRS) know, if it looks good.


jancami:
Looks good to me!

jdtsmith:
This looks very clean, thanks.  We should probably decide on a nomenclature for lines.  I think `name_lam` where `lam` is the wavelength to 0.1µm should suffice.  We'll have to be typing these in quite a bit to pull them out of tables, so the full spectroscopic terms are a bit cumbersome.  What do others think? 

I've also been thinking about how we name/brand science packs.  Since in principle our science packs are resolution-independent, we probably shouldn't name them/target-them for single instruments.  But them how to name them?  PAHFIT Classic is self-explanatory.

jdtsmith:
Also, for the leading comments, we should come up with some reasonable nomenclature and a pointer to the docs, like Jan did for the iso instrument pack.

els1:
Regarding the name of the file, I suggest `PDR_pack`.  The fact it is called 'PDR' indicates it is a science pack. Also, 'pack' and 'file' seems a duplicate.

Ameek-Sidhu:
Updated the yaml file. I have changed the names of the lines following the convention `name_lam` and have renamed the file to `PDR_pack.yml`.

drvdputt:
The PDR pack is invalid because there is a duplicate line: H2_Q(1) appears twice. Perhaps it should be H2_O(1)?

Ameek-Sidhu:
> The PDR pack is invalid because there is a duplicate line: H2_Q(1) appears twice. Perhaps it should be H2_O(1)?

There was an error in the line name. It is indeed H2_Q(1) line but for a different vibrational level. Since, we decided not to enter the names of vibrational levels, I was supposed to write the first two letters of the line in case of duplication, which I forgot in this case. I have it fixed it now though.

jdtsmith:
Closing, as I think this work should be subsumed into the `pdr.yaml` pack we've talked about.  @Ameek-Sidhu if you want to start that with your PDR_PACK, can you submit that as a standalone PR to the dev branch (though I'd call it simply pdr.yaml)?

It's also possible @drvdputt is already internally working with such a pack, so perhaps we can use a simple distillation between your two efforts to get that file started. Main differences from `gal.yaml` (also TBD) are line inclusivity, plus perhaps some wider allowance for PAH feature shape changes (TBD).

PR #217, opened 2022-06-27
fix S07 att; still keeping the hybrid form as opposed to a tabulated form 

ThomasSYLai:


els1 (approved):
Looks good.

jdtsmith:
@BethanyRS, does this PR fix your non-standard wavelength range fit issue?

jdtsmith (commented):
I'd recommend we break this out into it's own module, perhaps pahfit.model.attenuation.  For now it will just have kvt, but will allow us to add more as we go.  Ideally it makes as few assumptions about the underlying model as it can. @ThomasSYLai, thoughts?

ThomasSYLai:
> into

@jdtsmith Sure, this is a good approach when we have more extinction curves to explore. I'd propose we keep this current form for version 2.1 and we can build on top of this down the line in future versions.

jdtsmith:
Merging on @ThomasSYLai's report that the wavelength issue is fixed.

PR #219, opened 2022-06-28
Move base to model subpackage

jdtsmith:
This is the _start_ of a PR which will allow the YAML-based Features table to work with the existing `astropy.modeling` fitting functionality.  At the same time, we'll move the `initialize_model` and `fit_spectrum` methods from `helper.py` into `pahfit.model`.


jdtsmith:
@Ameek-Sidhu you may want to make your own PR for this, so I can push to it.  I don't think you can push to this PR branch, right?

PR #225, opened 2022-07-18
Area Gauss/Drude

Dhruvil2911:
In this PR, I've used area (power) as the parameter instead of the amplitude to fit Gaussian and Drude profiles.
I've made 2 classes AreaGaussian and AreaDrude in the component_models file corresponding to each profile.
There are corresponding changes in the helpers, base, and feature_strength files. I have also created an ipac file with the area columns (scipack_ExGal_area_1.ipac).
The command > run_pahfit M101_Nucleus_irs.ipac scipack_ExGal_area_1.ipac 
will demonstrate the changes.

For feature_strengths of Gaussian/Drude, I have simply equated it to the area (power).


pahfit/base.py
```diff
@@ -1,13 +1,15 @@
-from astropy.modeling.functional_models import Gaussian1D
+#from astropy.modeling.functional_models import Gaussian1D
```
karllark, line 1: Please delete if we are no longer going to use.

pahfit/base.py
```diff
@@ -1,13 +1,15 @@
...
     BlackBody1D,
     ModifiedBlackBody1D,
     S07_attenuation,
     att_Drude1D,
+    AreaGaussian1D,
+    AreaDrude1D
 )
 from astropy.table import Table, vstack
-from astropy.modeling.physical_models import Drude1D
+#from astropy.modeling.physical_models import Drude1D
```
karllark, line 12: Please delete if we are no longer going to use.

pahfit/base.py
karllark, line 1: Please delete if we are no longer going to use.

pahfit/component_models.py
```diff
@@ -134,3 +134,37 @@ def evaluate(x, tau, x_0, fwhm):
             profile = Drude1D(amplitude=1.0, fwhm=fwhm, x_0=x_0)
             tau_x = tau * profile(x)
             return (1.0 - np.exp(-1.0 * tau_x)) / tau_x
+
+
+class AreaGaussian1D(Fittable1DModel):
+
```
karllark, line 140: Please add a docstring to provide basic information about this function.  Like what currently exists for `S07_attenuation` for example.

pahfit/component_models.py
```diff
@@ -134,3 +134,37 @@ def evaluate(x, tau, x_0, fwhm):
...
+    def evaluate(x, area, mean, stddev):
+        """
+      AreaGaussian1D model function.
+      """
+        return ((1e14*area*(mean**2)/(np.sqrt(2*np.pi)*299792458*1e6*stddev)) * np.exp(
+          -0.5 * (x - mean) ** 2 / stddev ** 2))
+
+
+class AreaDrude1D(Fittable1DModel):
+
```
karllark, line 157: Docstring please.

karllark (changes requested):
Great to see this PR.

karllark:
Is the idea here to have pahfit only do line fitting based on area?  Or are we going to still allow central intensity?  If the former, then this PR is in good shape.  If the latter, then significant changes needed to support both.

The tests are failing as they are based on fitting the central intensity of lines.  This can be changes to use areas instead, but need to know the answer to my above queries to know the direction to go.

Dhruvil2911:
Thanks for reviewing the PR Karl. I've made the requested changes and added the docstring. Let me know if any other changes are required.

karllark:
> Is the idea here to have pahfit only do line fitting based on area? Or are we going to still allow central intensity? If the former, then this PR is in good shape. If the latter, then significant changes needed to support both.
> 
> The tests are failing as they are based on fitting the central intensity of lines. This can be changes to use areas instead, but need to know the answer to my above queries to know the direction to go.

Any response to this query?  @jdtsmith @els1

jdtsmith:
The new model will use power exclusively for lines and dust features.  So it’s a good idea IMO to get the current astropy model to as well.  I think @Ameek-Sidhu plans to just use amplitude for now as a stopgap.

jdtsmith:
> > The new model will use power exclusively for lines and dust features. So it’s a good idea IMO to get the current astropy model to as well. I think @Ameek-Sidhu plans to just use amplitude for now as a stopgap.
> 
> To be clear, this means we should change the tests to use power instead of amplitude, right? (Checking as I don't understand the "astropy model" in this context.) If so, this should be done in this PR and then all future work has to be done in power. I provide guidance on how to do this.

Yes, once implemented.  Actually all the tests need some attention I think (haven't looked closely at them).

karllark:
> Actually all the tests need some attention I think (haven't looked closely at them).

Tests always need more.  But they are working at this point until this change is made.  Hence the tests can be updated as part of this PR.  Then once it is merged, the tests will continue to work and do what we want - check that nothing has broken the core functionality.

jdtsmith:
Maybe my impressions are wrong but I've been getting test failure emails for months.  With apologies for my contributions to the increased entropy, I'm happy for any additional commits/PRs that help with the testing situation.  I'll leave this one to you and @Dhruvil2911; but ideally this PR would go in after @Ameek-Sidhu and @drvdputt get YAML-hookup + API cleanup.

karllark:
Briefly, the tests passed.  See https://github.com/PAHFIT/pahfit/actions.  ;-)  Takes vigilance to keep them passing.  BTW, I turned off email notification of failing tests long, long ago.  Just too many!

karllark:
Note that the tests are currently failing due to codestyle issues.  We have quite a suite of tests to make sure all parts of the repository work including the code, docs, codestyle, etc.

jdtsmith:
@karllark  @alexmaragko @drvdputt Do we intend to follow through with this?  No matter what happens with the new model, it seems smart to have an astropy.modeling version "along for the ride" for testing (and as a backup).  For that to work "for real", we need a power-based Gaussian and Drude.  Right now we are _lying_ and pretending amplitude = power.

karllark:
Yes this can be done fairly easily.

jdtsmith:
Great thanks @karllark ; would appreciate checking and cleaning up this PR (or generating a new one).  I believe we're close to merging the dev branch, so it could happen after that.   Then at least we could do our power-based fits.

jdtsmith:
We need to clarify the astropy-flavored pahfit model's _internal units_ to complete this.  This is TBD until dev merge.

jdtsmith:
This has been superseded by #295, continue conversation there.

PR #228, opened 2022-07-26
Add aperture-extracted spectra from MIRI MRS data of VV114

drvdputt:
The fringe correction is not perfect (still some wobbles at a scale of 4 wavelength points), but these might be usable.

What format should the included test spectra be in? What I did for these was just use the write function of Spectrum1D, and wrote them to fits (which is binary). Maybe text format (ipac like the rest) is preferred?

Caveats: 
- no uncertainties yet
- background subtraction not performed
- the residual fringe step was applied to the crf files. This paper https://arxiv.org/abs/2207.13067 applied it to the cal files (i.e. before combining the dithers), which is probably better.


drvdputt:
I figured something out to get uncertainties (basically quadratic summing the pixels). I added the aperture extraction code to my package here https://github.com/drvdputt/PAHFIT-cube/blob/master/pahfitcube/cube_aperture.py

jdtsmith:
Thanks Dries.  This looks great to me.  We should test a diversity of Spectrum1D-readable formats anyway.  Is there some segment identifying header keywords in the FITS files?  

Happy to commit later today unless others notice any issue.

jdtsmith:
Can we get an updated figure?

drvdputt:
Here are some plots. According to the uncertainties output by the JWST pipeline, the SNR is between 100 and 200 by the way (but if you zoom in on some parts, the noise is clearly bigger than that).

ch12

ch34

Here's a zoom-in of some early attempts at fitting these segments. Clearly still some fringing in certain locations, even with the residual fringe step applied.

Figure 8

jdtsmith:
These are great, thanks.  What do you make of the flux breaks?  Aperture-related?  On Monday Karl mentioned that for point sources the segment flux alignment is _perfect_.

jdtsmith:
Also could you post a snippet for loading these via Spectrum1D?

karllark:
Well perfect is a high ideal.  Very, very good.  The offsets shown are small.  Some may be improved with updates to the calibration reference files.  Very happy with how good this spectrum already looks!

drvdputt:
This should do for loading.

```
from pathlib import Path
from specutils import Spectrum1D
s1d_files = list(Path('./VV114-extracted/').glob('*s1d*'))
twelve_specs = [Spectrum1D.read(f) for f in s1d_files]
```

And something like this for using one of them in PAHFIT

```
from astropy.nddata import StdDevUncertainty

# convert Spectrum1D object to PAHFIT data format
if spec1d.uncertainty is None:
    # fake uncertainty if there is none
    unc = StdDevUncertainty(0.05 * spec1d.flux)
else:
    unc = spec1d.uncertainty
    obsdata = {"x": spec1d.spectral_axis.to(u.micron), "y": spec1d.flux, "unc": unc.quantity}
```

jdtsmith:
This looks great, thanks @drvdputt.  Merging for now, and we can always make improvements to these files when new calibrations/etc. appear.

PR #229, opened 2022-08-17
Multi + glob-style instrument specification for stitched spectra fitting

jdtsmith:
This supports fitting stitched spectra by allowing the user to specify an iterable of instruments instead of just a single value per input spectrum:

``` python
ins=("jwst.miri.mrs.ch1.A","jwst.miri.mrs.ch1.B")
```

and/or glob-style match strings:

``` python
ins="jwst.miri.mrs.ch*.[AB]"
```

It works by identifying all overlapping instrument segments at the given wavelengths, computing all relevant resolutions, and using the _average_ value with lower and upper bounds set to the min and max predicted resolution.  Lines which intersect only one segment have masked bounds (i.e., are fixed).

Note that, if multiple segments are provided or matched:

1. the return value of `wave_ranges()` can be a list of (2-element) lists.
2. `fwhm()` and `resolution()` return `(n, 3)` _masked_ arrays (given `n` input wavelengths).  The additional 2 values specify lower and upper bounds.

Also note that `instruments()` now takes an optional `match` argument, which can be used to test a given match:

``` python
In [399]: instruments(('jwst.nirspec.*.medium','jwst.miri.mrs.ch*.[AB]'))
Out[399]: 
['jwst.nirspec.g140.medium',
 'jwst.nirspec.g235.medium',
 'jwst.nirspec.g395.medium',
 'jwst.miri.mrs.ch1.A',
 'jwst.miri.mrs.ch1.B',
 'jwst.miri.mrs.ch2.A',
 'jwst.miri.mrs.ch2.B',
 'jwst.miri.mrs.ch3.A',
 'jwst.miri.mrs.ch3.B',
 'jwst.miri.mrs.ch4.A',
 'jwst.miri.mrs.ch4.B']
```

### DIFF
--- a/pahfit/errors.py
+++ b/pahfit/errors.py
@@ -4,3 +4,7 @@ class PAHFITFeatureError(Exception):
 
 class PAHFITPackError(Exception):
     pass
+
+
+class PAHFITWarning(Warning):
+    pass
--- a/pahfit/packs/instrument.py
+++ b/pahfit/packs/instrument.py
@@ -37,6 +37,16 @@ def read_instrument_packs():
     packs = dict(_ins_items('', packs))  # Flatten list
 
 
+def _ins_items(path, tree):
+    """Generator for traversing packs."""
+    if isinstance(tree, dict) and not tree.get("range"):
+        _dot = "." if len(path) > 0 else ""
+        for key, sub in tree.items():
+            yield from _ins_items(f"{path}{_dot}{key}", sub)
+    else:
+        yield path, tree
+
+
 def pack_element(segments):
     """Return the pack element(s) for the given segment name(s).
 
@@ -80,16 +90,6 @@ def pack_element(segments):
             ret.append(packs[s])
 
     return ret
-
-
-def _ins_items(path, tree):
-    """Generator for traversing packs."""
-    if isinstance(tree, dict) and not tree.get("range"):
-        _dot = "." if len(path) > 0 else ""
-        for key, sub in tree.items():
-            yield from _ins_items(f"{path}{_dot}{key}", sub)
-    else:
-        yield path, tree
 
 
 def instruments(match=None):
@@ -143,7 +143,7 @@ def resolution(segment, wave_micron):
     out[..., 0] = np.mean(res, 0)
 
     multi = np.count_nonzero(res, 0) > 1  # waves matching more than one segment
-    if np.count_nonzero(multi) > 0:
+    if np.count_nonzero(multi) > 0:  # add lower/upper bounds
         out[multi, 1] = np.min(res[:, multi], 0)
         out[multi, 2] = np.max(res[:, multi], 0)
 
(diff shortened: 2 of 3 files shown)


drvdputt:
I have merged this into my clone of the #222 branch.

Things are a bit fiddly because some functions return different things depending on whether there's one or multiple instrument segments specified. (wave_range and fwhm being the most important ones). Currently making some extra functions to work around this. Things like a wavelength range check, and a fwhm recommendation function that gives one number + fixed or not + upper and lower bounds (the latter are already returned by the current implementation).

jdtsmith:
Thanks; happy to simplify to make it easier to use.  I left the default case (one segment) returning just a normal array, and multi-segments loosening the bounds.  If you think it's better I can always return bounded `(n,3)` `fwhm`, etc., where the bounds will be masked for single segments.    

Also, thoughts on `wave_range`?   When there are multiple, I could "merge" them using min/max, but what if the user passes in non-continuous spectra, like `jwst.miri.mrs.*.[AB]`?

drvdputt:
I have worked around it for now. But I feel having a consistent behavior independent of the number of segments would be more maintainable.

As for wave_range, instead of returning the range explicitly, it might be more interesting to just have some functions that tell you what the deal is for a given wavelength.

For example I created this function
```
def test_waves_in_any_segment(wave_micron, segments):
    """Test if each given wavelength is covered by at least one instrument
    
    Parameters
    ----------
    wave_micron: dimensionless numpy array of wavelengths in micron
    segments: list of segments
    
    Returns
    -------
    np.array of bool, True where wave_micron[i] is is the range of any of the segments."""
```

For non-continuous spectra, a similar check could be done to remove any wavelength points that fall in between segments. The details of that can be discussed (do we want to stop everything and raise an error, or just discard the offending wavelength points and continue).

jdtsmith:
> it might be more interesting to just have some functions that tell you what the deal is for a given wavelength.

I was actually thinking this same thing. I think it should also have a keyword like `near` which, if enabled, computes the FWHM at any of the segment ends, and uses it to estimate if you are within a fixed multiple of `fwhm` (2.25 is what I've been using), either direction (basically dilates the wave_ranges by 2.25 FWHMs before the test).  I'm happy to adapt this for inclusion. 

> For non-continuous spectra, a similar check could be done to remove any wavelength points that fall in between segments. The details of that can be discussed (do we want to stop everything and raise an error, or just discard the offending wavelength points and continue).

I think don't worry about it.  If `segments` map to a non-continuous set, so be it, and we will only tell you if you are "in" (or "near") at least one of them.  You can then ask for the `fwhm` and be given a (possible bounded) value. Right now `fwhm` for multi-segments only returns values for "in".  Need to evaluate this for "near" as well.

jdtsmith:
Please checkout the new `within_segment` function to see how it works, and give it a check over.  For unifying the output, what do you think about adding an optional flag `as_bounded` that always returns bounded output.  This way interactive use is still convenient for the common case of "plot resolution vs. wavelength" etc.

jdtsmith:
I added a new `check_range` function that performs the "out of bounds" warning/error checks, as discussed in #230, which this addresses.

jdtsmith:
Note that check_range is never called automatically, so we should do so first thing upon fit/guess/etc.

jdtsmith:
Since instrument.py isn't yet used in master, I'm going to merge this and then cut the `dev:api+yaml` "trunk" branch.

PR #236, opened 2022-10-03
Fix nirspec resolution

### LINKED ISSUE #235: Instrument packs: Incorrect coefficients for NIRSpec resolution
The resolution returned by pahfit.instrument.resolution is not correct, at least not for nirspec. Looking at this plot https://jwst-docs.stsci.edu/files/97979440/97979452/1/1596073265647/jwst_nirspec_g140h_disp.png, from this page https://jwst-docs.stsci.edu/jwst-near-infrared-spectrograph/nirspec-instrumentation/nirspec-dispersers-and-filters#NIRSpecDispersersandFilters-DispersioncurvesfortheNIRSpecdispersers, it should be around 1500 at 0.8 micron, for G140H. Instead we get

```
>>> instrument.resolution('jwst.nirspec.g140.high', 0.8)
3490.42860832
```

These are the coefficients from the instrument pack:
```
            coefficients: [2291.3574, 1404.4167, 88.98144, 36.30806]
```
The zeroth order coefficient seems too high.

jdtsmith:
NISPEC polynomial fits to resolution had implicit 'domain' set from
numpy.polynomial.Polynomial.  The domain was not saved, so evaluating
the polynomial led to differet results.  All segments were re-fit with
default ([-1,1]) domain, except for prism, which did not converge
without wavelength rescaling due to unusual shape.  Direct support of
domain was added.

Please check against published resolution results. Closes #235.

### DIFF
--- a/pahfit/packs/instrument.py
+++ b/pahfit/packs/instrument.py
@@ -85,7 +85,7 @@ def pack_element(segments):
         for s in sm:
             if packs[s].get('polynomial') is None:
                 try:
-                    p = Polynomial(packs[s]['coefficients'])
+                    p = Polynomial(packs[s]['coefficients'], domain=packs[s].get('domain'))
                     packs[s]['polynomial'] = p
                     packs[s]['range_fwhm'] = [x / p(x) for x in packs[s]['range']]
                 except KeyError:
--- a/pahfit/packs/instrument.py
+++ b/pahfit/packs/instrument.py
@@ -85,7 +85,7 @@ def pack_element(segments):
         for s in sm:
             if packs[s].get('polynomial') is None:
                 try:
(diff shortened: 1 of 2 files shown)


drvdputt:
Did a quick test at a couple of wavelengths, for each nirspec high-resolution segment. All the returned resolution values are reasonable now, when compared to the plots on the webpage that I linked in the issue.

jdtsmith:
Thanks.  I realized with the `convert` method I could fit with auto-domain then convert to domain-less, so I could remove that complexity.  Mind taking another quick peek?

image

drvdputt:
I also checked the medium resolution just to be sure. It all seems correct now.

Issue #53, opened 2019-02-24
Units for model pack and output format


karllark:
Should add the units to the model pack and output files.  This would require the ingest of the model pack to make sure to convert the units to the internally used ones.  And probably have the output convert back to the input units for consistency.  

Should be fairly straightforward to implement, but including good check for robustness would be good and add complexity.

PR #241, opened 2022-11-02
Features: read support for `tied` parameters and group; and working "units" package

jdtsmith:
This PR (to dev branch) adds the ability to read `tied' features:

``` yaml
            power:
                tied:   # A one-to-many sum tie (varying ratio), line to DFs
                    feature: stable_dfs
                    ratio:
                        value: 1.25
                        bounds: [-0.25#, 0.25#]

```

`tied` can appear per feature parameter, or at the group level (for many-to-* ties).  

The tied constraint information is kept in `features.meta[_ratios]`.  Since the latter cannot serialize the `np.ma.masked` constant, these are set to `None` (e.g. for missing bounds).  Note that this is as of yet only read support; the current model does not yet support tied features.

### DIFF
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -25,7 +25,7 @@ import astropy.units as u
 from pkg_resources import resource_filename
 from pahfit.errors import PAHFITFeatureError
 from pahfit.features.features_format import BoundedMaskedColumn, BoundedParTableFormatter
-from pahfit.units import UNITS
+import pahfit.units
 
 # Feature kinds and associated parameters
 KIND_PARAMS = {'starlight': {'temperature', 'tau'},
@@ -36,9 +36,9 @@ KIND_PARAMS = {'starlight': {'temperature', 'tau'},
                'absorption': {'wavelength', 'fwhm', 'tau', 'geometry'}}
 
 # Parameter default units: flux density/intensity/power (other units determined on fit)
-PARAM_UNITS = {'temperature': UNITS.temperature.value,
-               'wavelength': UNITS.wavelength.value,
-               'fwhm': UNITS.wavelength.value}
+PARAM_UNITS = {'temperature': pahfit.units.temperature,
+               'wavelength': pahfit.units.wavelength,
+               'fwhm': pahfit.units.wavelength}
 
 
 class UniqueKeyLoader(yaml.SafeLoader):
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -25,6 +25,20 @@ import astropy.units as u
 from pkg_resources import resource_filename
 from pahfit.errors import PAHFITFeatureError
 from pahfit.features.features_format import BoundedMaskedColumn, BoundedParTableFormatter
+from pahfit.units import UNITS
+
+# Feature kinds and associated parameters
+KIND_PARAMS = {'starlight': {'temperature', 'tau'},
+               'dust_continuum': {'temperature', 'tau'},
+               'line': {'wavelength', 'power'},  # 'fwhm', Instrument Pack detail!
+               'dust_feature': {'wavelength', 'fwhm', 'power'},
+               'attenuation': {'model', 'tau', 'geometry'},
+               'absorption': {'wavelength', 'fwhm', 'tau', 'geometry'}}
+
+# Parameter default units: flux density/intensity/power (other units determined on fit)
+PARAM_UNITS = {'temperature': UNITS.temperature,
+               'wavelength': UNITS.wavelength,
+               'fwhm': UNITS.wavelength}
 
 
 class UniqueKeyLoader(yaml.SafeLoader):
@@ -114,7 +128,7 @@ class Features(Table):
 
     Note that each parameter has an associated `kind', and that each
     kind has an associated set of allowable parameters (see
-    `_kind_params`, below).
+    `KIND_PARAMS`.
 
     See Also
     --------
@@ -125,14 +139,6 @@ class Features(Table):
     MaskedColumn = BoundedMaskedColumn
 
     param_covar = TableAttribute(default=[])
-    _kind_params = {'starlight': {'temperature', 'tau'},
-                    'dust_continuum': {'temperature', 'tau'},
-                    'line': {'wavelength', 'power'},  # 'fwhm', Instrument Pack detail!
-                    'dust_feature': {'wavelength', 'fwhm', 'power'},
-                    'attenuation': {'model', 'tau', 'geometry'},
-                    'absorption': {'wavelength', 'fwhm', 'tau', 'geometry'}}
-
-    _units = {'temperature': u.K, 'wavelength': u.um, 'fwhm': u.um}
     _param_attrs = set(('value', 'bounds', 'tied'))  # params can have these attributes
     _group_attrs = set(('bounds', 'features', 'kind', 'tied'))  # group-level attributes
     _no_bounds = set(('name', 'group', 'geometry', 'model'))  # String attributes (no bounds)
@@ -209,7 +215,7 @@ class Features(Table):
                 raise PAHFITFeatureError(f"No kind found for {name}\n\t{file}")
 
             try:
-                valid_params = cls._kind_params[kind]
+                valid_params = KIND_PARAMS[kind]
             except KeyError:
                 raise PAHFITFeatureError(f"Unknown kind {kind} for {name}\n\t{file}")
             unknown_params = [x for x in keys
@@ -351,7 +357,7 @@ class Features(Table):
         t[kind][name]['group'] = group
         t[kind][name]['kind'] = kind
         for (param, val) in pars.items():
-            if param not in cls._kind_params[kind]:
+            if param not in KIND_PARAMS[kind]:
                 continue
             if isinstance(val, dict) and 'tied' in val:
                 tie = val['tied']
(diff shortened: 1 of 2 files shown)


jdtsmith:
We can leave this for after dev merge.

jdtsmith:
I'll retarget this for main once dev is merged.

drvdputt:
I would be OK with merging this, IF the code in this pull request was actually in a working state.

There seems to be something wrong with tox (for all pull requests), so the tests didn't run here. But I ran them offline, and almost all of them fail with the same error. Here's the error of one of them.

```python
pytest
====================================== test session starts =======================================
... output bla bla ...
pahfit/tests/test_feature_parsing.py:37: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pahfit/features/features.py:181: in read
    return cls._read_scipack(file)
pahfit/features/features.py:298: in _read_scipack
    return cls._construct_table(feat_tables)
pahfit/features/features.py:428: in _construct_table
    col.unit = PARAM_UNITS[cn]
../../opt/miniconda3/envs/pahfit/lib/python3.10/site-packages/astropy/table/column.py:1184: in __setattr__
    super().__setattr__(item, value)
../../opt/miniconda3/envs/pahfit/lib/python3.10/site-packages/astropy/table/column.py:930: in unit
    self._unit = Unit(unit, parse_strict='silent')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <class 'astropy.units.core.Unit'>, s = <UNITS.temperature: Unit("K")>, represents = None
format = None, namespace = None, doc = None, parse_strict = 'silent'

    def __call__(self, s="", represents=None, format=None, namespace=None,
                 doc=None, parse_strict='raise'):
    
        # Short-circuit if we're already a unit
        if hasattr(s, '_get_physical_type_id'):
            return s
    
        # turn possible Quantity input for s or represents into a Unit
        from .quantity import Quantity
    
        if isinstance(represents, Quantity):
            if is_effectively_unity(represents.value):
                represents = represents.unit
            else:
                represents = CompositeUnit(represents.value *
                                           represents.unit.scale,
                                           bases=represents.unit.bases,
                                           powers=represents.unit.powers,
                                           _error_check=False)
    
        if isinstance(s, Quantity):
            if is_effectively_unity(s.value):
                s = s.unit
            else:
                s = CompositeUnit(s.value * s.unit.scale,
                                  bases=s.unit.bases,
                                  powers=s.unit.powers,
                                  _error_check=False)
    
        # now decide what we really need to do; define derived Unit?
        if isinstance(represents, UnitBase):
            # This has the effect of calling the real __new__ and
            # __init__ on the Unit class.
            return super().__call__(
                s, represents, format=format, namespace=namespace, doc=doc)
    
        # or interpret a Quantity (now became unit), string or number?
        if isinstance(s, UnitBase):
            return s
    
        elif isinstance(s, (bytes, str)):
            if len(s.strip()) == 0:
                # Return the NULL unit
                return dimensionless_unscaled
    
            if format is None:
                format = unit_format.Generic
    
            f = unit_format.get_format(format)
            if isinstance(s, bytes):
                s = s.decode('ascii')
    
            try:
                return f.parse(s)
            except NotImplementedError:
                raise
            except Exception as e:
                if parse_strict == 'silent':
                    pass
                else:
                    # Deliberately not issubclass here. Subclasses
                    # should use their name.
                    if f is not unit_format.Generic:
                        format_clause = f.name + ' '
                    else:
                        format_clause = ''
                    msg = ("'{}' did not parse as {}unit: {} "
                           "If this is meant to be a custom unit, "
                           "define it with 'u.def_unit'. To have it "
                           "recognized inside a file reader or other code, "
                           "enable it with 'u.add_enabled_units'. "
                           "For details, see "
                           "https://docs.astropy.org/en/latest/units/combining_and_defining.html"
                           .format(s, format_clause, str(e)))
                    if parse_strict == 'raise':
                        raise ValueError(msg)
                    elif parse_strict == 'warn':
                        warnings.warn(msg, UnitsWarning)
                    else:
                        raise ValueError("'parse_strict' must be 'warn', "
                                         "'raise' or 'silent'")
                return UnrecognizedUnit(s)
    
        elif isinstance(s, (int, float, np.floating, np.integer)):
            return CompositeUnit(s, [], [], _error_check=False)
    
        elif isinstance(s, tuple):
            from .structured import StructuredUnit
            return StructuredUnit(s)
    
        elif s is None:
            raise TypeError("None is not a valid Unit")
    
        else:
>           raise TypeError(f"{s} can not be converted to a Unit")
E           TypeError: UNITS.temperature can not be converted to a Unit

../../opt/miniconda3/envs/pahfit/lib/python3.10/site-packages/astropy/units/core.py:2059: TypeError
```

drvdputt:
I also expect more errors to pop up because now the columns suddenly have units. Might need to do `.value` in a couple of places in the `Model` code.

jdtsmith:
Well what I'm thinking is that (pure, but impure for now too) `Model` will:

1. Determine if we are doing surface brightness or flux density, ala (no mixing!  -> error):
   ```
   uparts = sp.flux.unit.decompose()
   isSB = (u.rad, -2) in zip(uparts.bases, uparts.powers)
   ```
2. Convert all incoming Spectrum1D's internally to the correct working unit (using your `new_flux_unit()` discovery).  
2. Record the correct "working unit" from units.py in the appropriate `Features` table columns. 

All/both our `ModelCore` backends will naturally expect to work in the working `UNITS`, so there will be no ambiguity of conversion for them to do themselves.  The features table will be updated with OUR units.  And we will document how to convert any given table/sum/etc. into other units.

jdtsmith:
The bug above was because Enum's seem dumb: you have to say Class.member.value to get at the actual "thing being held".  Happy to switch to something else if you have an idea.

jdtsmith:
OK, duh.  Too much C programming in my background.  Module constants it is.  BTW, do you have an idea why all the tests are failing?  I keep seeing:

```
py38-alldeps: failed with pass_env values cannot contain whitespace, use comma to have multiple values in a single line, invalid values found 'HOME WINDIR LC_ALL LC_CTYPE CC CI TRAVIS'
```

drvdputt:
## tox issue

I found the problem. It has something to do with tox, because the error is this
```
py38-alldeps: failed with pass_env values cannot contain whitespace, use comma to have multiple values in a single line, invalid values found 'HOME WINDIR LC_ALL LC_CTYPE CC CI TRAVIS'
```
and tox.ini has this
```
# Pass through the following environment variables which may be needed for the CI
passenv = HOME WINDIR LC_ALL LC_CTYPE CC CI TRAVIS
```

I found this thread from 5 days ago https://github.com/tox-dev/tox/issues/2676

Apparently a breaking change going from tox 3 to tox 4 https://tox.wiki/en/latest/faq.html#tox-4-changed-ini-rules, and the continuous integration is indeed installing tox 4:
```
Collecting tox
  Downloading tox-4.0.11-py3-none-any.whl (142 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 142.1/142.1 kB 4.3 MB/s eta 0:00:00
```

Even though the documentation states passenv=SPACE-SEPARATED-GLOBNAMES [link](https://tox.wiki/en/3.1.0/config.html#confval-passenv=SPACE-SEPARATED-GLOBNAMES), they were actually supposed to be comma or newline separated 

## PAHFIT tests

I tested this branch using the tests (running locally) and the demo notebook, and everything seems fine. The new table header looks like this when saved with the embedded units.
```
# %ECSV 1.0
# ---
# datatype:
# - {name: name, datatype: string}
# - {name: group, datatype: string}
# - {name: kind, datatype: string}
# - {name: temperature, unit: K, datatype: string, format: 0.4g, subtype: 'int64[3]'}
# - {name: tau, datatype: string, format: 0.4g, subtype: 'float64[3]'}
# - {name: wavelength, unit: um, datatype: string, format: 0.4g, subtype: 'float64[3]'}
# - {name: power, datatype: string, format: 0.4g, subtype: 'float64[3]'}
# - {name: fwhm, unit: um, datatype: string, format: 0.4g, subtype: 'float64[3]'}
# - {name: model, datatype: string}
# - {name: geometry, datatype: string}
```

So far no issues, because the old code in `base.py` does `.data` on every column, and therefore ignores whether there are units or not.

karllark:
For the tox issue - adding commas between the variables in passenv fixes the issue.  Just tested this in another repository.

jdtsmith:
You guys are too good. The SB vs flux unit dichotomy reminds me: if a model features table is in SB and the user tries to reuse that model to fit other spectra with flux units (or vv), we should error.

jdtsmith:
@drvdputt can you submit a quick tox fix to get past this error?  Let's get a plan together to merge dev to main soon; especially important since that's what you demoed last week.

drvdputt:
I've just posted #253. Once that's merged, you can rebase this pull request branch onto it, and we'll see if the tests work.

jdtsmith:
We can hold off on this until dev hits main.  I'll also separate the `tied` and "default units" stuff into two PRs so we can evaluate the latter for issues.   We do need to clarify the units situation, since our science packs have some implicit units already (e.g. microns).  

@drvdputt are there any particular units-related issues you think the "default unit" portion of this PR this will hit?  If so, perhaps discuss them in #53.

PR #242, opened 2022-11-07
New guess function

drvdputt:
I have replaced the old guess function by one that lives in Model. (old code in PAHFITBase is still there for reference, but no longer used).

Main features:
- Does not construct a model. Instead, directly writes guess result to features table.
- Only guesses non-fixed values
- Blackbody continuum guess is almost same algorithm as original guess function. Difference: starlight guess uses  flux(wmin) / BB(5 micron) if wmin > 5, instead of flux(5.5 micron) / BB(5 micron). Has to do with this https://github.com/PAHFIT/pahfit/pull/167
- Line powers are guessed as: amplitude = power_guess / fwhm, with power_guess the trapezoidal integral over a 4 x FWHM window (window size was chosen rather randomly, and could be improved)
- Dust features do not have a guess yet. I can try to come up with something while I'm tweaking the line guess a bit more.


drvdputt:
I made some more changes, and here is an illustration

With the M101 data: the guess
Figure 1(3)
The fit:
Figure 4

With NIRSPEC data
Guess:
Figure 40
Fit:
Figure 42

pahfit/model.py
```diff
@@ -166,11 +170,111 @@ def guess(self, spec: Spectrum1D, redshift=None):
...
+            return amp_guess
+
+        loop_over_non_fixed("starlight", "tau", starlight_guess)
+
+        # count number of blackbodies in the model
+        nbb = len(self.features[self.features["kind"] == "dust_continuum"])
+
+        def dust_continuum_guess(row):
+            temp = row["temperature"][0]
+            fmax_lam = 2898.0 / temp
```
jdtsmith, line 211: I don't know that the max wavelength is ideal here, since it may be well off the coverage area of the input spectra.

pahfit/model.py
```diff
@@ -166,11 +170,111 @@ def guess(self, spec: Spectrum1D, redshift=None):
...
+            bb = BlackBody1D(1, temp)
+            if fmax_lam >= wmin and fmax_lam <= wmax:
+                w = fmax_lam
+                flux_ref = sp(w)
+            elif fmax_lam > wmax:
+                w = wmax
```
jdtsmith, line 217: But I suppose this works well enough.

pahfit/model.py
```diff
@@ -166,11 +170,111 @@ def guess(self, spec: Spectrum1D, redshift=None):
...
+                flux_ref = yz[np.argmax(xz)]
```
jdtsmith, line 218: Why not interpolate the flux by doing a median of points near the chosen wavelength `w`?

pahfit/model.py
```diff
@@ -166,11 +170,111 @@ def guess(self, spec: Spectrum1D, redshift=None):
...
+
+        def amp_guess(row, fwhm):
+            w = row["wavelength"][0]
+            if not instrument.within_segment(w, inst):
+                return 0
+
+            factor = 1.5
+            wmin = w - factor * fwhm
+            wmax = w + factor * fwhm
+            xz_window = np.logical_and(wmin < xz, xz < wmax)
```
jdtsmith, line 244: I usually prefer this like `(xz > wmin) & (xz < wmax)`.

pahfit/model.py
```diff
@@ -166,11 +170,111 @@ def guess(self, spec: Spectrum1D, redshift=None):
...
+            xpoints = xz[xz_window]
+            ypoints = yz[xz_window]
+            if np.count_nonzero(xz_window) >= 2:
+                # difference between flux in window and flux around it
+                power_guess = integrate.trapezoid(yz[xz_window], xz[xz_window])
```
jdtsmith, line 249: Not sure if this may be too heavy.  But people doing lots of fits can of course avoid guessing (or use a simpler, guess(simple_scale = True) thing we develop later).

pahfit/model.py
```diff
@@ -166,11 +170,111 @@ def guess(self, spec: Spectrum1D, redshift=None):
...
+                # subtract continuum estimate, but make sure we don't go negative
+                continuum = (ypoints[0] + ypoints[-1]) / 2 * (xpoints[-1] - xpoints[0])
+                if continuum < power_guess:
+                    power_guess -= continuum
+            else:
+                power_guess = 0
+
+            return power_guess / fwhm
```
jdtsmith, line 257: We are (approximately) returning to amplitude here because we don't yet "actually have power"?

pahfit/model.py
```diff
@@ -166,11 +170,111 @@ def guess(self, spec: Spectrum1D, redshift=None):
...
+            loop_over_non_fixed("line", "fwhm", line_fwhm_guess, force=True)
+
+        def df_amp_guess(row):
+            w = row["wavelength"][0]
+            fwhm=row["fwhm"][0]
+            if not (instrument.within_segment(w - fwhm, inst) or instrument.within_segment(w + fwhm, inst)):
+                return 0
+
+            drude = Drude1D(amplitude=1, x_0=w, fwhm=fwhm)
+            return sp(w) / drude(w)
```
jdtsmith, line 274: This seems a bit overly complex.  E.g. why not just evaluate at `x_0` (where the Drude will peak)?

jdtsmith (changes requested):
This looks like a real improvement to the simple guesser.  Probably more we can do to get ourselves close.  For continuum (DF and starlight), some "bottom quartiling" might help find those low points to hit.

jdtsmith:
> old code in PAHFITBase is still there for reference

Is the intention to remove this file soon?  (Git will always remember).

pahfit/model.py
drvdputt: Yes. I wrote it in terms of power so that it's easy to adjust later, once the model works with power units.

pahfit/model.py
drvdputt: I see, Drude1D evaluated at x_0 is always equal to amplitude. Will change when we have power-based drudes though.

jdtsmith:
Is this one necessary prior to dev->main merge, @drvdputt ?  I think we discussed how this might be a bit brittle in the face of very different instrument packs, assuming "continuum adjacent to lines", etc.  So I feel like it needs more testing.  Or is this the very simple guess, translated to interact with the `Features` table?

drvdputt:
The only part that *may be needed* and works relatively well, is the better guess for the lines, to get them to the right order of magnitude. 
An added bonus, is that with this new function, we can remove the old guess code from base.py

On the other hand, this will need to change when we have power units. And the dust feature guesses are too sketchy right now, they often result in bad fits for the continuum.

So either
1. We close this, and I redo this code when we get the units working.
2. I trim it down, add only the lines part, and remove the old guess code.

drvdputt:
> Or is this the very simple guess, translated to interact with the Features table?

This is an added bonus; we can remove the old guess code from base.py, which interacts with `param_info`, and instead, we interact directly with with the Features table.

jdtsmith:
Thanks.  I think it's worth having the "simple dumb" guess which addresses the Features table in this, since part of our api is now `model.guess(...)`.  So if this can be massaged into that, I'd vote to include it in the dev merge.

drvdputt:
I've just rebased this on the dev branch and solved the conflicts.

Changes added
1. Removed "smart" dust guess, and replaced it by the old 0.5 * median guess. I think it would be a good idea to do the same for the lines amplitude guess, but a keyword to activate the "smart line guess" would be nice. I found it useful for very bright lines.
2. Removed old guess from base.py
3. Removed `_backport_param_info()`, which was for converting old guess output to features table

I will do some testing of this function on Monday.

drvdputt:
There is yet another issue that mentions the problem with the old guess function when the wavelengths of the input data don't have the right range (#262). Based on all the questions I've gotten so far, this is by far the most common bug that people struggle with, so we really need to get this merged. I'll do a quick review of my code soon (and you comments, @jdtsmith).

pahfit/model.py
drvdputt: I replaced this with the old method (setting amplitude equal to median flux), since trying to be smart makes the fitter prone to getting stuck in local minima.

pahfit/model.py
drvdputt: Trying to hit "low points" in the spectrum might indeed work better. Still, I'm not having too much trouble with the way it works now.

pahfit/model.py
drvdputt: looks like I put in this change some time ago

drvdputt:
I did a quick check, and most of the comments were already addressed; the code snippets in this thread were just outdated, and I just never replied that I changed things.

I've posted some replies, and resolved everything. This function seems to be working well for my work with the Orion & NGC7023 NIRSpec data.

One more small comment from experience: instead of doing a "smart" guess for the line fluxes, I found that simply calling "model.fit()" twice actually works really well to get the lines "picked up", in cases where only a few lines were actually fit properly. Seems like resetting the minimization from where it left off can really help with some issues.

jdtsmith:
Can you comment more on your `model.fit` point?  Do you mean you restart the fit at the same parameter values it resulted in during the first fit?  To me this sounds like insufficient convergence latitude for the optimizer.  Do you think starting line amplitudes closer to their final value is helpful?  I know you had a fancier line estimator before (that may not be robust).

drvdputt:
The "integrate_line_fluxes" option I added definitely helps when the lines peak is several orders of magnitude higher than the other flux values. But sometimes, not all lines are picked up the right way, even with this estimate. 

With "fitting twice", I mean using the output of the first fit, as the initial conditions for the second one. This is something I have been doing in my own work, but not very elegant of course. A solution with tweaked fit parameters (or a different algorithm?) would of course be preferable if performance matters.

drvdputt:
In any case, the "fitting twice works better"-phenomenon can be addressed in a separate issue. I will likely experiment with other fitters while I'm dealing with new spectra in the near future.

The new guess algorithm might still have some issues, but I think we should merge this PR to get the mechanical changes in place. After that, tweaks to the algorithm should be straightforward (whenever we find something that works better.

And equally important: the new guess solves the long-standing wavelength range issue (e.g. https://github.com/PAHFIT/pahfit/issues/260#issuecomment-1516600646)

jdtsmith:
OK.  When we get the `Fitter` abstract class implemented we'll have to refactor this, but sounds good.

PR #244, opened 2022-11-21
read_spectrum as a wrapper around Spectrum1D.read

drvdputt:
I found that the specutils reader doesn't care if the columns are named 'sigma' or 'stddev'. So just by changing the reader to Spectrum1D, we can now read in all our included spectra.

I also added a test for this read-in step. It was already useful to iron out some bugs and convince myself that this will work for everything we need. Should be useful in the future too, when new example spectra get added.

### DIFF
--- a/pahfit/helpers.py
+++ b/pahfit/helpers.py
@@ -81,7 +81,7 @@ def read_spectrum(specfile, format=None):
     if format is None:
         suffix = specfile.split(".")[-1].lower()
         if suffix == "ecsv":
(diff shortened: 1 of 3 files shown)


drvdputt:
Thanks to this test, I found another bug: the data files in the subdirectories of pahfit/data are not installed! 

The solution is to put these in setup.cfg too (I don't think recursive glob syntax is supported by setup.cfg). Adding a commit that does this.

PR #261, opened 2023-06-07
Update iso.yaml to fix AOT1 ranges

sarduval:
Previous ranges (Version 0) did not fully cover the actual wavelength range of ISO SWS, as stated in Table 1 and Table 8 of: https://iopscience.iop.org/article/10.1086/375443/pdf . This was mainly an issue for AOT1, which has an overall range of [2.35-45.39] micron. Updated ranges (Version 1), as of June 2023.


jdtsmith:
Thanks for looking at this @sarduval.  Can you add a note in the header of this file mentioning the change and this URL?

PR #253, opened 2023-01-04
Fix tox run

drvdputt:
Fixes #251

PR #269, opened 2023-12-14
updating for astropy6

### LINKED ISSUE #268: Can't import pahfit since astropy 6.0.0
In the astropy template, which was used to set up the pahfit package, there is the file `pahfit/_astropy_init.py`. There is some code in there which does no longer work with astropy 6.0.0, because `update_default_config` was removed from `astropy.config.configuration`.

```python
...

if not _ASTROPY_SETUP_:  # noqa
    import os
    from warnings import warn
    from astropy.config.configuration import (
        update_default_config,
        ConfigurationDefaultMissingError,
        ConfigurationDefaultMissingWarning,
    )

    # Create the test function for self test
    from astropy.tests.runner import TestRunner

    test = TestRunner.make_test_runner_in(os.path.dirname(__file__))
    test.__test__ = False
    __all__ += ["test"]

    # add these here so we only need to cleanup the namespace at the end
    config_dir = None

    if not os.environ.get("ASTROPY_SKIP_CONFIG_UPDATE", False):
        config_dir = os.path.dirname(__file__)
        config_template = os.path.join(config_dir, __package__ + ".cfg")
        if os.path.isfile(config_template):
            try:
                update_default_config(__package__, config_dir, version=__version__)
            except TypeError as orig_error:
                try:
                    update_default_config(__package__, config_dir)
                except ConfigurationDefaultMissingError as e:
                    wmsg = (
                        e.args[0] + " Cannot install default profile. If you are "
                        "importing from source, this is expected."
                    )
                    warn(ConfigurationDefaultMissingWarning(wmsg))
                    del e
                except Exception:
                    raise orig_error

   ...
```

We need to look at this code and see if there's a trivial change we can do. And then set an appropriate astropy version constraint.

As a workaround, I just commented things out everything referring to update_default_config.

### LINKED ISSUE #271: Readthedocs build failing
`build.os` has to be set explicitly now in readthedocs.yaml

You will find a few examples here https://github.com/readthedocs/readthedocs.org/issues/8912, in the linked issues and pull requests.

karllark:
Updates for astropy 6.0.  Includes fixing the URL for the astropy developer docs.

Also updating the readthedocs and tox configurations.

Closes #268 
Closes #271


.github/workflows/tex-tests.yml
```diff
@@ -27,16 +27,16 @@ jobs:
         # For example -- os: [ubuntu-latest, macos-latest, windows-latest]
         os: [ubuntu-latest, macos-latest, windows-latest]
 
-        # Test python 3.8, 3.9, and 3.10 by default.
-        python-ver: [8, 9, 10]
+        # Test python version 3.x
+        python-ver: [9, 10, 11]
```
drvdputt, line 31: Should we add 12? It was released in october.

.github/workflows/tex-tests.yml
```diff
@@ -27,16 +27,16 @@ jobs:
...
 
         # Specify which tox environments to test in this list.
         # tox-env: [cov, alldeps, devdeps, astropylts]
         tox-env: [alldeps]
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v1
```
drvdputt, line 37: Why is this being downgraded to v1? Would upgrading to v4 (the current recommendation by github) suffice too?

.github/workflows/tex-tests.yml
```diff
@@ -27,16 +27,16 @@ jobs:
...
     - name: Set up python 3.${{ matrix.python-ver }} with tox environment py3${{ matrix.python-ver }}-${{ matrix.tox-env }} on ${{ matrix.os }}
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v1
```
drvdputt, line 39: Same here

.github/workflows/tex-tests.yml
drvdputt: See this pull request where I upgrade these to fix a deprecation warning https://github.com/PAHFIT/pahfit/pull/253

pahfit/_astropy_init.py
drvdputt, line 1: So you removed this file. Can you provide a quick comment on what it was doing exactly?

tox.ini
drvdputt, line 1: I agree that this dependency coverage was probably overkill. As long as everything works with the newest version of everything, we should be good. Just recommend that people have an up to date environment. Then we just need to catch bugs as new versions arrive.

drvdputt (changes requested):
General comment: there quite a few changes in this, and probably only a few are relevant to fixing the bug. While I think it would be ok, a brief explanation for the extra changes might be nice.

Changes still needed: if you change the actions versions to v1 again, I think another deprecation warning will pop up, since github actions no longer runs on Node 12, but Node 16 instead. The old versions might not be compatible anymore with whatever version of Node github is using now under the hood. I recommend upgrading everything to v4, and seeing if it works.

.github/workflows/tex-tests.yml
karllark: I had some issues with tests with 12 in other repositories.  So maybe delay for now.

.github/workflows/tex-tests.yml
karllark: Will do.

pahfit/_astropy_init.py
karllark: No longer needed.   That is one of the updates.

karllark:
Yes, ready to merge.  So I will do such.

Issue #257, opened 2023-03-04
New internal API: agnostic functional model setup / results retrieval


jdtsmith:

drvdputt:
Just a reminder that doing this will help us to finally remove base.py. The old features-to-model converter can be moved to a `Fitter` subclass, and the plotting code can be moved into `Model`.

jdtsmith:
This is the highest priority action.  The one thing I'm questioning is whether the fitter cores (`Fitter` subclasses, e.g. `APFitter` and `SPFitter` for astropy and scipy fitters) should "know about" and directly interact with `Features` tables.  

The advantage of having them do so is it's less work: you can pass a `Features` table directly in to e.g. `fit`, and it will do the right thing and update it _in place_ (and attach the covariance matrix if available).

The disadvantage is now you are quite strongly coupled, so any future changes to `Features` would need to preserve backward compatibility.

One thing is clear is that `Fitter`'s should not need to know _anything_ about instruments.

Maybe it helps to think about _what_ kind of thing `Fitter.SomeSubClass.fit` would accept and return, if not a `Features` table?  The IDL MPFIT equivalent is a list of structures with parameter name, starting value, boundedness, and bounds; the so called `parinfo` structure.  Somebody actually ported this to python; [here's their parinfo details](https://github.com/segasai/astrolibpy/blob/ee9b260b14eea8ec593ba067bff9ee5092abc0ff/mpfit/mpfit.py#L184).  

Another detail, to save room for future improvements:

Soon we'd like to fit _collections_ of `Spectrum1D` objects.  My fitter's approach is to concat these into one big "franken-spectrum" with non-monotonic wavelengths, and then specify segment constraints.  So e.g. a narrow line will apply to one segment, and a wider (lower-resolution) copy will apply to another.  Once we dilate the width of Drude's too, this will happen commonly.  

It would be easy enough to just hand `fit` a list of `Spectrum1D`'s (suitably tagged with instrument name), and let it set up its fit internally for that.  I don't know if AstroPy will get there, but we should save room for it to do so via the JointFitter.  In some ways this argues for letting each `Fitter` sub-class know about and read/interact with `Features` tables directly.  

Thoughts?

drvdputt:
So, I started writing down my thoughts, and it became a bit long. See a little summary at the end.

[...]

PR #270, opened 2024-01-23
Abstract fitter API draft

drvdputt:
As discussed in #257 I abstracted out parts from my new model builder branch. This abstract API describes what I need to keep the current astropy fitting implementation working.

@jdtsmith Let's see if my idea is compatible with your new fitter.

Note that only the fitter.py file was added. This API is not used by any other code yet. `Model` is still using the astropy fitting framework directly. All of the astropy compound model machinery should be moved to a subclass of `Fitter` called `APFitter`, after which Model will be simplified, while preserving the original behavior.

### DIFF
--- a/pahfit/fitter.py
+++ b/pahfit/fitter.py
@@ -111,7 +111,7 @@ class Fitter(ABC):
         pass
 
     @abstractmethod
-    def register_dust_features(self, name, power, wavelength, fwhm):
+    def register_dust_feature(self, name, power, wavelength, fwhm):
         """Register a dust feature.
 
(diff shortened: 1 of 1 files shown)


jdtsmith:
Great, thanks Dries, very interested to iterate on this with you.  I'm dreadfully busy for two and half weeks.  I'll be in Baltimore early next month maybe we could find 30min to sit down with it.

jdtsmith:
Is this superseded by #289 and therefore should be closed?

drvdputt:
Yes, that is correct. I can close it myself.

PR #278, opened 2024-04-19
Use importlib.resources instead of pkg_resources (from setuptools)

### LINKED ISSUE #277: Use of pkg_resources
In pahfit/features/features.py, `pkg_resources` is used. In a few cases (e.g. freshly created environment with Python venv and then pip install pahfit), I have run into an import error for this. The quick fix is always to `pip install setuptools`. For a smoother install experience, we should either

1. Include setuptools as a dependency in setup.cfg
2. Refactor the code to use the native `importlib.resources` instead (https://docs.python.org/3/library/importlib.resources.html)
(I'm not sure if pkg_resources is deprecated or not)

drvdputt:
This seems to work fine to resolve #277 . Let's see what the tests do.

### DIFF
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -22,7 +22,7 @@ import numpy as np
 from astropy.table import vstack, Table, TableAttribute
 from astropy.io.misc.yaml import yaml
 import astropy.units as u
-from pkg_resources import resource_filename
+from importlib import resources
 from pahfit.errors import PAHFITFeatureError
 from pahfit.features.features_format import BoundedMaskedColumn, BoundedParTableFormatter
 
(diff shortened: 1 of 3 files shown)


drvdputt:
In addition to the pull requests planned in #280 , can we pull this one in first?

jdtsmith:
I know very little about packaging, but happy to merge this.  Can you target this to the `dev` branch, which I just created as a clean branch from master?

drvdputt:
The issue here is that we were using a packaging tool (always had to `pip install setuptools`) to deal with the import of our resource files (the example spectra and included science packs). 

This was never a problem with the automated tests, since they seem to always have setuptools in their environment. This is why we never included setuptools as a dependency. So when setting up a virtual environment with `python3 -m venv venv_name`, it is not always guaranteed that setuptools is present.

With these changes, we use the native `importlib`, so one less dependency. I think we can merge this into dev. If any thing breaks, I'm sure all the other testing during development will reveal it.

jdtsmith:
Great done. I'm pretty ignorant on all the flavors of `setup*` stuff.  I can at least pip install a local directory though ;).

Issue #274, opened 2024-01-30
features.util.bounded_is_fixed() does not work for dust_feature-only science packs


drvdputt:
When a science pack is created that only has dust_feature entries, there will be no masked values in the power column of the Features table. As a consequence, the astropy table constructor automatically decides whether to use a `Column` instead of a `MaskedColumn`. This means that any functionality that relies on the presence of a mask attribute, wil fail.

E.g. in `featuresutil.bounded_is_fixed(): return ma.getmask(val)[..., -2:].all(-1)`, where `getmask` will return `False`.

Most consistent solution: Force the use of `MaskedColumn` (or in our case `BoundedMaskedColumn`) when the Features table.

Even if we found another way around this bug, I think we should enforce this to ensure consistency. (e.g. What if the user want to set power boundaries, but the power column does not support masks because of the way the table was loaded?)

jdtsmith:
I agree we can just enforce all columns are `BoundedMaskedColumns` even if most of the elements are "fixed".

jdtsmith:
OK, update, never mind that.  #283 includes a fix for this.  It should create bounds for all parameter columns except for string-parameters (`name`, `group`, `model`, `geometry`).  The latter should never be passed to the bounds checker functions.

drvdputt:
Yes, seems to work now, since `BoundedMaskedColumn` doesn't exist anymore, this bug can also no longer appear.

jdtsmith:
I didn't check all the `utils` convenience functions but we can revisit if needed.

Issue #273, opened 2024-01-30
Obscure bug in Features table formatter


drvdputt:
In `Features`, the `TableFormatter` class is overridden by a new class definition: `BoundedParTableFormatter` from  `features_format.py`. It overrides the `_pformat_table` function, to pretty-print the 3-tuples representing (value, min, max) in each BoundedMaskedColumn of the table. 

It does this by setting col.info.format to a custom formatting function, which assumes that it will receive a 3-tuple. 
This happens in this code block

```python

    colsh = [(col, col.shape) for col in table.columns.values()]
            BoundedMaskedColumn._omit_shape = True
            for col, sh in colsh:
                if len(sh) == 2 and sh[1] == 3:
                    bpcols.append((col, col.info.format))
                    col.info.format = fmt_func(col.info.format or "g")

```

after this, the super()._pformat_table is invoked to apply these formatters.

I'm still not sure how to reproduce  this, but printing the features table throws an exception sometimes. For some reason the formatting function `_fmt` receives a scalar value instead of a 3-tuple. This happens during the assignment `col.info.format = ...`. Astropy `Column` uses a custom assigment function for this property, that tries to format a test value. I think there is some sort of automatic unpacking going on in that test function, that results in an element of the tuple being passed, instead of the intended actual 3-tuple in the column. I have no solution for this, except guarding for this in our `_fmt` function.

Here's the series of exceptions
```pdb
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
File ~/Projects/horsehead+ncg7023/analyze-spectra/git/venv/lib/python3.11/site-packages/astropy/table/pprint.py:90, in get_auto_format_func.<locals>._auto_format_func(format_, val)
     89 try:
---> 90     out = format_func(format_, val)
     91     if not isinstance(out, str):

File ~/Projects/horsehead+ncg7023/analyze-spectra/git/venv/lib/python3.11/site-packages/astropy/table/pprint.py:88, in get_auto_format_func.<locals>._auto_format_func.<locals>.<lambda>(format_, val)
     87 if callable(format_):
---> 88     format_func = lambda format_, val: format_(val)
     89     try:

File ~/Repositories/pahfit/pahfit/features/features_format.py:16, in fmt_func.<locals>._fmt(v)
     15 print(f"problem in _fmt: v = {v}, fmt = {fmt}, type(fmt) = {type(fmt)}")
---> 16 raise e

[...]

Issue #266, opened 2023-11-21
Crash in astropy.table.pprint when using lines-only science pack


drvdputt:
# Bug description

The code crashes when trying to print Features.

## Reproducing

This is a strange bug, reproducible with very minimal examples

This crashes:
```python
from pahfit.model import Model
m = Model.from_yaml('minimal_bad.yaml')
print(m)
```
with the following minimal yaml file
```yaml
H2_lines:
    kind: line
    wavelength:
        H2_S(8):    5.0530
```

On the other hand, if a stellar component is added, everything is fine
```
starlight:
    kind: starlight
    temperature: 5000

H2_lines:
    kind: line
    wavelength:
        H2_S(8):    5.0530
```

## Exceptions and tracebacks

 <details>
  <summary>
    Click for full output
  </summary>

Traceback (most recent call last):
  File "/Users/dvandeputte/opt/miniconda3/envs/pahfitcube/lib/python3.11/site-packages/astropy/table/pprint.py", line 90, in _auto_format_func
    out = format_func(format_, val)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dvandeputte/opt/miniconda3/envs/pahfitcube/lib/python3.11/site-packages/astropy/table/pprint.py", line 88, in <lambda>
    format_func = lambda format_, val: format_(val)
                                       ^^^^^^^^^^^^
  File "/Users/dvandeputte/Repositories/pahfit/pahfit/features/features_format.py", line 16, in _fmt
    raise e
  File "/Users/dvandeputte/Repositories/pahfit/pahfit/features/features_format.py", line 9, in _fmt
    if ma.is_masked(v) or ma.is_masked(v[0]):
                                       ~^^^
IndexError: invalid index to scalar variable.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/dvandeputte/opt/miniconda3/envs/pahfitcube/lib/python3.11/site-packages/astropy/table/pprint.py", line 535, in _pformat_col_iter
    yield format_col_str(idx)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/dvandeputte/opt/miniconda3/envs/pahfitcube/lib/python3.11/site-packages/astropy/table/pprint.py", line 523, in format_col_str
    left = format_func(col_format, col[(idx,) + multidim0])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #259, opened 2023-03-17
Standardize units

jdtsmith:
## What 
This introduces a new file `units.py` which standardizes the units for science-packs, and eventually for the entire `Features`->`Model`->`Fitter` chain. 

## Why

This will:

1. make a proper AREA-based Gaussian/Drude fitter possible (see #225)
2. standardize the report of value in results `Features` tables
3. ensure most values are within say 5 dex of unity for table glance-ability and to helping with underflow.  

**Note**: we will _convert_ any user input units into this format, and optionally convert tabulated/plotted output to the user's spectral units.  But for specifying the _PAHFIT science model_ to be fitted itself, these are the working units [1]. 

See [this comment](https://github.com/PAHFIT/pahfit/issues/28#issuecomment-1349676272) for information on how we can leverage `astropy.units` to understand user-input spectral units, e.g. to discriminate between flux density and spectral intensity (aka surface brightness).

## Proposed PAHFIT working + output units summary:

| Type | Working Unit | 
| ----- | ------------- |
| Wavelength/FWHM/spectral_axis[2] | µm | 
| Temperature[2]| K |
| Flux density | mJy | 
| Flux power | 1e-22 W/m<sup>2</sup> | 
| — OR — |
| Surface Brightness/Intensity | MJy/sr | 
| Surface Brightness Power | 1e-10 W/m<sup>2</sup>/sr |

[1] Internally, fitters may adopt any different units for convenience or other reasons, but that is a hidden implementation detail.  For the user, these will be the working units.
[2] These are already effectively standards in our science packs

PR #282, opened 2024-05-07
Workarounds for issues with Features column definition and formatting

drvdputt:
There were a series of strange issues with the Features class in certain edge cases
- There are only dust features (#274)
- There are only lines (#266)
- Exceptions are thrown due to a mismatch in the number of elements in a table entry (#273)

To solve #274, both the Column and MaskedColumn definitions of Features were set to BoundedMaskedColumn.
The formatting hacks for Features were revised to deal with the above change and with #273 and #266.

The new formatting method, takes the default output of the formatter (using if/else to cover all possible strings created for a 3-tuple), and modifies it through string replacements.

This pull request is independent from #281, so one can go in after the other (at most a simple rebase required).

### DIFF
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -110,6 +110,7 @@ class Features(Table):
 
     TableFormatter = BoundedParTableFormatter
     MaskedColumn = BoundedMaskedColumn
+    Column = BoundedMaskedColumn
 
     param_covar = TableAttribute(default=[])
     _kind_params = {'starlight': {'temperature', 'tau'},
--- a/pahfit/features/features_format.py
+++ b/pahfit/features/features_format.py
@@ -6,11 +6,16 @@ from astropy.table.pprint import TableFormatter
 # * Special table formatting for bounded (val, min, max) values
 def fmt_func(fmt):
     def _fmt(v):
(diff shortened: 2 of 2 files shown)


jdtsmith:
A very relevant conversation about the pretty-printing is [here](https://astropy.slack.com/archives/C067V74GK/p1651713325599399).  You can see me dealing with the fact that the default printer omits the middle of a 3-tuple (like our bounded param).

jdtsmith:
I _think_ this is entirely superseded by the much simpler #283; please have a look at that.  Here's my proposal: let's merge 

1. #259 
2. #281 (once review is addressed)
3. #283 (pending your review) 

to `dev`, test that combo.  If all's well on `dev` proceed with the next steps (targeting `dev` until we're happy).

drvdputt:
Ok, I will review the above three pull requests. I'll post another comment here once I'm done with this, and then we'll pull those into dev. I will then test test some of my edge cases in dev, to see if we still need any extra hacks on top of #283.

jdtsmith:
I think this has been superseded by #283 and can be closed, yes?

drvdputt:
Yes, we have a different approach in the other pull request now. Closing.

PR #283, opened 2024-05-09
Simplify feature bounds using np.nan instead of masking

jdtsmith:
Partially masked arrays are not really well supported, and required elaborate workarounds for pretty printing.  Here we switch to using a _structured_ array type (`val`, `min`, `max` fields),  and represent "fixed" bounds with a pair of `np.nan` values for `min` and `max`,  _instead of_ masking them.  I.e. `nan` is interpreted as the same as negative or positive infinity, unless both bounds are `nan`, which indicates a fixed parameter.   

"Fully open" (unconstrained) bounds are expressed just as before, with `(-np.inf, np.inf)`.  But in general: don't do those ;).

### DIFF
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -69,8 +69,8 @@ def value_bounds(val, bounds):
     Returns:
     -------
 
-      The value, if unbounded, or a 3 element tuple (value, min, max).
-        Any missing bound is replaced with the numpy `masked' value.
+      A 3 element tuple (value, min, max).
+        Any missing bound is replaced with the numpy.nan value.
 
     Raises:
     -------
@@ -82,7 +82,7 @@ def value_bounds(val, bounds):
     if val is None:
         val = np.ma.masked
     if not bounds:
-        return (val,) + 2 * (np.nan,)  # Fixed
+        return (val,) + 2 * (np.nan,)  # (val,nan,nan) indicates fixed
     ret = [val]
     for i, b in enumerate(bounds):
         if isinstance(b, str):
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -82,7 +82,7 @@ def value_bounds(val, bounds):
     if val is None:
         val = np.ma.masked
     if not bounds:
-        return (val,) + 2 * (np.ma.masked,)  # Fixed
+        return (val,) + 2 * (np.nan,)  # Fixed
     ret = [val]
     for i, b in enumerate(bounds):
         if isinstance(b, str):
@@ -123,7 +123,8 @@ class Features(Table):
     _group_attrs = set(('bounds', 'features', 'kind'))  # group-level attributes
     _param_attrs = set(('value', 'bounds'))  # Each parameter can have these attributes
     _no_bounds = set(('name', 'group', 'geometry', 'model'))  # String attributes (no bounds)
-
+    _bounds_dtype = np.dtype([("val", "f4"), ("min", "f4"), ("max", "f4")])
+    
     @classmethod
     def read(cls, file, *args, **kwargs):
         """Read a table from file.
@@ -308,7 +309,7 @@ class Features(Table):
         tables = []
         for (kind, features) in inp.items():
             kind_params = cls._kind_params[kind]  # All params for this kind
-            rows = []
+            rows, dtypes = [], []
             for (name, params) in features.items():
                 for missing in kind_params - params.keys():
                     if missing in cls._no_bounds:
@@ -316,8 +317,8 @@ class Features(Table):
                     else:
                         params[missing] = value_bounds(0.0, bounds=(0.0, None))
                 rows.append(dict(name=name, **params))
-            table_columns = rows[0].keys()
-            t = cls(rows, names=table_columns)
+                dtypes.append(None if name in cls._no_bounds else cls._bounds_dtypes)
+            t = cls(rows, names=rows[0].keys(), dtype=dtypes)
             for p in cls._kind_params[kind]:
                 if p not in cls._no_bounds:
                     t[p].info.format = "0.4g"  # Nice format (customized by Formatter)
@@ -352,7 +353,7 @@ class Features(Table):
                 pass
             else:
                 # mask only the value, not the bounds
-                row[col_name].mask[0] = mask_value
+                row[col_name].mask['val'] = mask_value
 
     def unmask_feature(self, name):
         """Remove the mask for all parameters of a feature."""
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -355,3 +355,7 @@ class Features(Table):
     def unmask_feature(self, name):
         """Remove the mask for all parameters of a feature."""
         self.mask_feature(name, mask_value=False)
+
+    def _base_repr_(self, *args, **kwargs):
+        """Omit dtype on self-print."""
+        return super()._base_repr_(*args, ** kwargs | dict(show_dtype=False))
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -316,7 +316,8 @@ class Features(Table):
                     else:
                         params[missing] = value_bounds(0.0, bounds=(0.0, None))
                 rows.append(dict(name=name, **params))
-                dtypes.append(None if name in cls._no_bounds else cls._bounds_dtype)
+            dtypes = [None if x in cls._no_bounds else cls._bounds_dtype for x in kind_params]
+            dtypes.insert(0, None)  # For the name
             t = cls(rows, names=rows[0].keys(), dtype=dtypes)
             for p in cls._kind_params[kind]:
                 if p not in cls._no_bounds:
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -24,7 +24,7 @@ from astropy.io.misc.yaml import yaml
 import astropy.units as u
 from pkg_resources import resource_filename
 from pahfit.errors import PAHFITFeatureError
-from pahfit.features.features_format import BoundedMaskedColumn, BoundedParTableFormatter
+from pahfit.features.features_format import BoundedParTableFormatter
 
 
 class UniqueKeyLoader(yaml.SafeLoader):
@@ -109,7 +109,6 @@ class Features(Table):
     """
 
     TableFormatter = BoundedParTableFormatter
-    MaskedColumn = BoundedMaskedColumn
 
     param_covar = TableAttribute(default=[])
     _kind_params = {'starlight': {'temperature', 'tau'},
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -123,7 +123,7 @@ class Features(Table):
     _param_attrs = set(('value', 'bounds'))  # Each parameter can have these attributes
     _no_bounds = set(('name', 'group', 'geometry', 'model'))  # String attributes (no bounds)
     _bounds_dtype = np.dtype([("val", "f4"), ("min", "f4"), ("max", "f4")])
-    
+
     @classmethod
     def read(cls, file, *args, **kwargs):
         """Read a table from file.
@@ -316,7 +316,7 @@ class Features(Table):
                     else:
                         params[missing] = value_bounds(0.0, bounds=(0.0, None))
                 rows.append(dict(name=name, **params))
-                dtypes.append(None if name in cls._no_bounds else cls._bounds_dtypes)
+                dtypes.append(None if name in cls._no_bounds else cls._bounds_dtype)
             t = cls(rows, names=rows[0].keys(), dtype=dtypes)
             for p in cls._kind_params[kind]:
                 if p not in cls._no_bounds:
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -121,7 +121,7 @@ class Features(Table):
     _units = {'temperature': u.K, 'wavelength': u.um, 'fwhm': u.um}
     _group_attrs = set(('bounds', 'features', 'kind'))  # group-level attributes
     _param_attrs = set(('value', 'bounds'))  # Each parameter can have these attributes
-    _no_bounds = set(('name', 'group', 'geometry', 'model'))  # String attributes (no bounds)
+    _no_bounds = set(('name', 'group', 'kind', 'geometry', 'model'))  # str attributes (no bounds)
     _bounds_dtype = np.dtype([("val", "f4"), ("min", "f4"), ("max", "f4")])
 
     @classmethod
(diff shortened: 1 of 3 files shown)


pahfit/features/features.py
```diff
@@ -308,16 +308,16 @@ def _construct_table(cls, inp: dict):
...
                 for missing in kind_params - params.keys():
                     if missing in cls._no_bounds:
                         params[missing] = 0.0
                     else:
                         params[missing] = value_bounds(0.0, bounds=(0.0, None))
                 rows.append(dict(name=name, **params))
-            table_columns = rows[0].keys()
-            t = cls(rows, names=table_columns)
+                dtypes.append(None if name in cls._no_bounds else cls._bounds_dtypes)
+            t = cls(rows, names=rows[0].keys(), dtype=dtypes)
```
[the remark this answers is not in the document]
drvdputt: I changed some things. Is this the intended result:
```
rows:  [{'name': 'starlight', 'group': '_none_', 'kind': 'starlight', 'temperature': (5000, nan, nan), 'tau': (0.0, 0.0, inf)}]
names:  dict_keys(['name', 'group', 'kind', 'temperature', 'tau'])
dtype:  [dtype('<U'), dtype('<U'), dtype('<U'), dtype([('val', '<f4'), ('min', '<f4'), ('max', '<f4')]), dtype([('val', '<f4'), ('min', '<f4'), ('max', '<f4')])]
resulting t.dtype [('name', '<U9'), ('group', '<U6'), ('kind', '<U9'), ('temperature', [('val', '<f4'), ('min', '<f4'), ('max', '<f4')]), ('tau', [('val', '<f4'), ('min', '<f4'), ('max', '<f4')])]
```

pahfit/features/features.py
```diff
@@ -82,7 +82,7 @@ def value_bounds(val, bounds):
     if val is None:
         val = np.ma.masked
     if not bounds:
-        return (val,) + 2 * (np.ma.masked,)  # Fixed
+        return (val,) + 2 * (np.nan,)  # (val,nan,nan) indicates fixed
```
drvdputt, line 85: You said that a single np.nan would represent infinity, but both np.nan represents fixed. How are [-inf, +inf] bounds represented then? I assume -np.inf, +np.inf? 

Am I correct to interpret this as?: we should never have a single np.nan in there, but if it happens, interpret it as np.inf as a failsafe

pahfit/features/features.py
```diff
@@ -308,16 +308,16 @@ def _construct_table(cls, inp: dict):
...
                     else:
                         params[missing] = value_bounds(0.0, bounds=(0.0, None))
                 rows.append(dict(name=name, **params))
-            table_columns = rows[0].keys()
-            t = cls(rows, names=table_columns)
+                dtypes.append(None if name in cls._no_bounds else cls._bounds_dtypes)
+            t = cls(rows, names=rows[0].keys(), dtype=dtypes)
             for p in cls._kind_params[kind]:
                 if p not in cls._no_bounds:
                     t[p].info.format = "0.4g"  # Nice format (customized by Formatter)
```
drvdputt, line 323: With my changes explaiend above applied, we're getting somewhere. But `Features.read(packfile)` is crashing on this line. When "0.4g" is assigned to t[p].info.format, the astropy code will perform a test of the assigned formatting function. This ultimately results in the following errors (multiple, because some are caught after which another is raised)
```
E               ValueError: unable to parse format string 0.4g for its column.
E                   ValueError: Unable to parse format string "0.4g" for entry "(0., 0., inf)" in column "tau"
E           ValueError: Invalid format for column 'tau': could not display values in this column using this format
```

pahfit/features/features.py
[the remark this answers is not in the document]
drvdputt: To set up this dtype, I use
```python
kind_cols = rows[0].keys()
col_dtypes = [np.dtype(type(rows[0][col_name])
                          if col_name in cls._no_bounds
                          else cls._bounds_dtype
                          for col_name in kind_cols]
```

pahfit/features/features.py
drvdputt: I removed this line (replaced by `pass`) to see what would happen. The code now continues, and here's a quick test
```python
f = Features.read('classic.yaml')
# change one of the bounds. I like this more explicit syntax.
f.loc['PAH_5.3']['fwhm']['max'] = 0.3 
f.pprint()
```

The output is 
```
In [8]: f.pprint()
    name       group          kind      temperature [val, min, max] ... fwhm [val, min, max]      model      geometry
                                                     K              ...          um                                  
----------- ------------ -------------- --------------------------- ... -------------------- --------------- --------
  starlight       _none_      starlight                5000 (fixed) ...                   --              --       --
dust_cont00    dust_cont dust_continuum                 300 (fixed) ...                   --              --       --
dust_cont01    dust_cont dust_continuum                 200 (fixed) ...                   --              --       --
dust_cont02    dust_cont dust_continuum                 135 (fixed) ...                   --              --       --
dust_cont03    dust_cont dust_continuum                  90 (fixed) ...                   --              --       --
dust_cont04    dust_cont dust_continuum                  65 (fixed) ...                   --              --       --
dust_cont05    dust_cont dust_continuum                  50 (fixed) ...                   --              --       --
dust_cont06    dust_cont dust_continuum                  40 (fixed) ...                   --              --       --
dust_cont07    dust_cont dust_continuum                  35 (fixed) ...                   --              --       --
    H2_S(7)     H2_lines           line                          -- ...                   --              --       --
    H2_S(6)     H2_lines           line                          -- ...                   --              --       --
    H2_S(5)     H2_lines           line                          -- ...                   --              --       --
    H2_S(4)     H2_lines           line                          -- ...                   --              --       --
    H2_S(3)     H2_lines           line                          -- ...                   --              --       --
    H2_S(2)     H2_lines           line                          -- ...                   --              --       --
    H2_S(1)     H2_lines           line                          -- ...                   --              --       --
    H2_S(0)     H2_lines           line                          -- ...                   --              --       --
     [ArII]  ionic_lines           line                          -- ...                   --              --       --
    [ArIII]  ionic_lines           line                          -- ...                   --              --       --
      [SIV]  ionic_lines           line                          -- ...                   --              --       --
     [NeII]  ionic_lines           line                          -- ...                   --              --       --
    [NeIII]  ionic_lines           line                          -- ...                   --              --       --
  [SIII]_18  ionic_lines           line                          -- ...                   --              --       --
      [OIV]  ionic_lines           line                          -- ...                   --              --       --
  [FeII]_26  ionic_lines           line                          -- ...                   --              --       --
  [SIII]_33  ionic_lines           line                          -- ...                   --              --       --
     [SiII]  ionic_lines           line                          -- ...                   --              --       --
  [FeII]_35  ionic_lines           line                          -- ...                   --              --       --
    PAH_5.3       _none_   dust_feature                          -- ...    0.17918 (-∞, 0.3)              --       --
    PAH_5.7       _none_   dust_feature                          -- ...       0.1995 (fixed)              --       --
    PAH_6.2       _none_   dust_feature                          -- ...       0.1866 (fixed)              --       --
    PAH_6.7       _none_   dust_feature                          -- ...       0.4683 (fixed)              --       --
   PAH_7.7a  PAH_7.7_cmp   dust_feature                          -- ...      0.93492 (fixed)              --       --
   PAH_7.7b  PAH_7.7_cmp   dust_feature                          -- ...       0.3344 (fixed)              --       --
   PAH_7.7c  PAH_7.7_cmp   dust_feature                          -- ...      0.41605 (fixed)              --       --
    PAH_8.3       _none_   dust_feature                          -- ...       0.4165 (fixed)              --       --
    PAH_8.6       _none_   dust_feature                          -- ...      0.33579 (fixed)              --       --
   PAH_10.7       _none_   dust_feature                          -- ...       0.2136 (fixed)              --       --
  PAH_11.3a PAH_11.3_cmp   dust_feature                          -- ...      0.13476 (fixed)              --       --
  PAH_11.3b PAH_11.3_cmp   dust_feature                          -- ...      0.36256 (fixed)              --       --
     PAH_12       _none_   dust_feature                          -- ...      0.53955 (fixed)              --       --
  PAH_12.6a PAH_12.6_cmp   dust_feature                          -- ...      0.53004 (fixed)              --       --
  PAH_12.6b PAH_12.6_cmp   dust_feature                          -- ...      0.16497 (fixed)              --       --
  PAH_13.48       _none_   dust_feature                          -- ...       0.5392 (fixed)              --       --
  PAH_14.04       _none_   dust_feature                          -- ...      0.22464 (fixed)              --       --
  PAH_14.19       _none_   dust_feature                          -- ...      0.35475 (fixed)              --       --
   PAH_15.9       _none_   dust_feature                          -- ...        0.318 (fixed)              --       --
    PAH_17a   PAH_17_cmp   dust_feature                          -- ...       0.2303 (fixed)              --       --
    PAH_17b   PAH_17_cmp   dust_feature                          -- ...       1.1076 (fixed)              --       --
    PAH_17c   PAH_17_cmp   dust_feature                          -- ...       0.2085 (fixed)              --       --
    PAH_17d   PAH_17_cmp   dust_feature                          -- ...      0.28592 (fixed)              --       --
   PAH_33.1       _none_   dust_feature                          -- ...        1.655 (fixed)              --       --
   silicate       _none_    attenuation                          -- ...                   -- S07_attenuation    mixed
```

drvdputt (changes requested):
I fully support the main concept of this pull request. The ('val' 'min' 'max') datatype feels nice and explicit. Much better than the cryptic 3-tuples from before (where you always had to index on 0, 1, or 2 to addres the needed value)

There are some mistakes, but I fixed them myself to see how the code would continue.

I got it working up to where the Features table is loaded. But the new data type and way of addressing the values, makes it so the code in `base.py` no longer works. I /could/ try to fix the base.py code. But would very much like to avoid this, as we will remove it soon...

pahfit/features/features.py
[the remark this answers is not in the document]
jdtsmith: Something else is wrong then.  Note the Table docs:

> The values must be valid [numpy.dtype](https://numpy.org/doc/stable/reference/generated/numpy.dtype.html#numpy.dtype) initializers or [None](https://docs.python.org/3/library/constants.html#None). Any list elements with value [None](https://docs.python.org/3/library/constants.html#None) fall back to the default type.

So we should be handing a list with some `None`'s in it.  Can you try again with the variable fix above?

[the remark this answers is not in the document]
drvdputt: I did try after doing that fix

This revealed two separate issues:
- the number of dtypes is not the same as the number of column names
names = ['name', 'group', 'kind', 'temperature', 'tau'] (5 values)
[dtype([('val', '<f4'), ('min', '<f4'), ('max', '<f4')])] (only one)
- when I manually pass `(None, None, None, dtype([('val', '<f4'), ('min', '<f4'), ('max', '<f4')]), dtype([('val', '<f4'), ('min', '<f4'), ('max', '<f4')]))`, then np.float32 is used as the default type, and the code will crash because those columns are actually strings.

pahfit/features/features.py
jdtsmith: Yeah that was dumb to leave in.  All `no_bounds` are simply strings, so leave it alone and `Table` will sort that out.

jdtsmith: Wait never-mind.  The idea is that the custom formatter shims this column format (which the user can change) in at the last second prior to printing.  But I guess this is being defeated by this testing, which happens outside the printer.  Sigh.  I want the user to be able to easily customize `col.info.format="%5.2f"` or whatever.  And to not worry about removing the function from that slot when we save (which results in an error).

jdtsmith: Maybe that's too cute.  Maybe we should just set up `col.info.format = fmt_func` right from the get-go, then fix up `Features.write` to replace it "just in time".  If you leave a function in that slot, you get:

```python
RepresenterError: ('cannot represent an object', <function _fmt at 0x118bedbc0>)
```

We could tell the user to alter `col.info.format2` or something if we want to use that in our formatter.  Thoughts?

drvdputt: Having an extra attribute for the "individual number" format seems like the most reliable way. Our _pformat_table can then grab this value to generate the right "val min max" formatting function with fmt_func(col.info.format2). 

Do we need a more informative name as opposed to format2? Maybe something that makes it more clear that it's a modification by pahfit?

pahfit/features/features.py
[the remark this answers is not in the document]
jdtsmith: OK dumb of me.  `name` is the name of the feature, not a parameter name.  I sent up a fix which should work, and now just hard-code `str` for the type of `_no_bounds` things.  I don't love that, because we might add a thing without (possible) bounds that is not a string, but OK for now.

pahfit/features/features.py
drvdputt: The use case I'm seeing here: its set to (val,nan,nan), and the user only changes one bound e.g. bla['max'] = 1.0. I think this will happen relatively commonly.

pahfit/features/features.py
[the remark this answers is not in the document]
drvdputt: More issues:
- `kind_params` does not include the generic columns 'name', 'group' and 'kind'. So the loop `dtypes = [str if x in cls._no_bounds else cls._bounds_dtype for x in kind_params]` misses those extra strings.
- `no_bounds` does not include 'kind', so this logic would give it the wrong type.

pahfit/features/features.py
jdtsmith: OK I have stamped out the bugs here I think.  I just used `str` for all the `no_bounds` stuff (adding `kind`).

pahfit/features/features.py
jdtsmith: In fact your example below (just altering the `'max'`) does exactly this, so it would be natural!

jdtsmith:
While working on the representation, I also stamped out the `dtype` printing on `__repr__` (which is oddly on by default), and allow now a `meta['pahfit_format']` on the table and/or individual columns, to affect the printed representation.  E.g.

```python
f.meta['pahfit_format']='.2f'
f['wavelength'].meta['pahfit_format']='.3f'
```

yielding:

```
...
    H2_S(7)     H2_lines           line              --             --  5.511 (fixed) 0.00 (0.00, ∞)              --              --       --
    H2_S(6)     H2_lines           line              --             --  6.109 (fixed) 0.00 (0.00, ∞)              --              --       --
    H2_S(5)     H2_lines           line              --             --  6.909 (fixed) 0.00 (0.00, ∞)              --              --       --
    H2_S(4)     H2_lines           line              --             --  8.026 (fixed) 0.00 (0.00, ∞)              --              --       --
...
```

To be documented somewhere.

drvdputt:
It looks like all of this is working as intended. At least until the actual until the Features -> astropy model part.

We cannot really make all the tests succeed for this, without revising all the pieces of code that directly address the Features table (I tried to make things work for a bit, but there are more of these than I expected). Especially, I cannot justify  revising all of base.py just to make the tests work, since I aim to delete it soon, when replaced by APFitter.

If we merge this, then we will have to continue in this broken state until I have integrated and adapted APFitter.

I suggest we merge the units pull request first (the extended version I just suggested, #284 )

pahfit/features/features.py
jdtsmith: Update: I do now allow `meta['pahfit_format']` on both table and columns.  Works great.

jdtsmith:
This sounds good.  What type of updates to APFitter do you mean?  `Features` is still accessible in the same way, and in fact `p[0]` and `p['val']` both work.

drvdputt:
Things that do not work anymore
- access like this: f['fwhm'][:, 0] now has to be f['fwhm']['val']. So I expect there to be more cases where not all the ways of accessing still work. 
- access to the `.mask` attribute does not seem to work in some cases

jdtsmith:
OK.  Overall I think the structured array is much cleaner.  I intend to merge your extend units PR #284 once finalized, then #281, then this #283.

jdtsmith:
Also do not worry about soon-to-be-forgotten `base.py`.

Issue #280, opened 2024-05-03
Piecewise/partial integration of drvdputt/experimental


drvdputt:
[github.com/drvdputt/pahfit/tree/experimental ](https://github.com/drvdputt/pahfit/tree/experimental) has diverged quite far from master. But it is in working order, and I am actively using it for my science goals. There are many aspects which we want to integrate into the master branch, but we should do so piece wise to avoid breakage.

Here is an overview of the changes I made so far.

*Major features or structures*
- Implemented Fitter abstract base class, and APFitter subclass. All astropy-fitting-specific code was moved into this class. See #270 for more details.
- Rewrite of model.plot, which makes use of tabulate() and (indirectly) the Fitter API where possible (current version crashes when no attenuation is in the model). Remove old plotting code from base.py.
- Use of the Fitter API in the Model class. Model building is much more straightforward. Removed old building code from base.py.
- Complete removal of pahfit.base. All functionality is split between Model and Fitter (and APFitter).
- Internal unit system. The file proposed in #259 was included, and used in various locations. Particularly important for the Power features above. *caveat*: So far, I only support MJy / sr or equivalent units (intensity). It is checked that the user-provided spectrum is compatible with the right physical quantity, and then the conversion is made internally. I did not find an obvious way to switch the Power feature classes dynamically based on the given unit type, so not flux units (Jy) yet.
- PowerGaussian and PowerDrude implementation (like #225, but unit conversions are much more clearly formulated, as the internal unit system can be used)
- Empty draft for Asymmetric features (just to test adding an asymmetry parameter to the YAML file)

*Other bug / usability fixes*
- Better exclusion algorithm for the features. Takes into account both the instrument model and the given wavelength range of the data. Having too many features outside of the data range makes the fitting unreliable (too many parameters with zero effect on the chi2).
- Always use "BoundedMaskedColumn" in the features table. This works around #266 (and #273, which are of the same origin)
- Rewrote the pretty print function of BoundedParTableFormatter to address #273. Was needed in combination with the above fix.
- Tried out different fitting algorithms. Slightly "better" results with TRFLSQFitter (less sticky to the bounds), but it seems more prone to getting stuck in local minima. Running both TRFLSQFitter and LevMarLSQFitter (one after the other) might provide the best of both worlds.
- Automatic documentation generation for the Model class.

jdtsmith:
Would be great to pull these into 4 PRs from prompt merging to main:

0. Internal `Fitter` API: 
1. Use API:
[...]

Issue #288, opened 2024-05-13
Implement Sara Profile


jdtsmith:
@sarduval has explored `Gaussian x Drude` convolutions (the functional form of which we nickname "the Sara profile".  She's come up with a mapping scheme to update a given Drude to another broader Drude, which works for most reasonable ranges of resolution and PAH features.  Right now our "parent" model for Drudes bakes in the instrument line spread function convolution: not good.  

The idea is that science packs should be "pure" and devoid of any artifacts of the instrument.  So just as we express lines as Gaussians of a computed FWHM, we'd compute updated FWHM for Drude's too.  Other features (BB, modifiedBB) are too broad for this to matter much.  

@drvdputt, any thoughts on where this should go?  I guess the `Fitter` is in charge of marrying instrument pack, figuring out gaussians, etc.?  So it should do this too.  Do we update `fwhm` for lines in the `Features` table after a fit?  We need a way of taking a fitted FWHM and running the transform backwards: so we can recover the true intrinsic feature width.

jdtsmith:
Relevant comment from PR#289, regarding redshift:

> When we expand Drude profile widths ala https://github.com/PAHFIT/pahfit/issues/288, we'll need to do something similar with the teleported line profile width, i.e. providing "the width of the line profile at the observed center of the feature, cast into the rest frame". We should probably clearly spell this out in a comment here.

PR #289, opened 2024-05-14
Fitter API and make test cases work

drvdputt:
Here next step for #280. Aside from the change in the features table introduced by #283, the user-facing `Model` things are basically unchanged (unless users were "digging for the astropy model" to do certain things).

For deep debugging or testing, one can still get  direct access to the astropy stuff via `model.fitter.model` (`model` is a `Model`, `model.fitter` is an `APFitter`, model.fitter.model is the underlying astropy model object).

You can look at the implementation details of `APFitter` to get a feel for how it works. But it is more important to take a good look at `Fitter`, and how its functions are used  in `Model`.

I also addressed the issues with the test cases (avoiding the use of `.loc`), and they seem to be running smoothly.

Besides complete removal of `base.py`, a reference to PAHFITBase in the documentation was also removed.

After this, only one more big change is needed: PowerGaussian and PowerDrude, and the change from amplitude to power units for the `Features` table.


pahfit/apfitter.py
jdtsmith, line 1: Can we create a new directory `pahfit/fitters` to place the fitters?

pahfit/features/features.py
```diff
@@ -137,7 +137,7 @@ class Features(Table):
     _group_attrs = set(('bounds', 'features', 'kind'))  # group-level attributes
     _param_attrs = set(('value', 'bounds'))  # Each parameter can have these attributes
     _no_bounds = set(('name', 'group', 'kind', 'geometry', 'model'))  # str attributes (no bounds)
-    _bounds_dtype = np.dtype([("val", "f4"), ("min", "f4"), ("max", "f4")]) # bounded param type
+    _bounds_dtype = np.dtype([("val", float), ("min", float), ("max", float)])  # bounded param type
```
jdtsmith, line 140: This reminds me we need to test `f4` vs. `f8` at some point.  I suppose that could be a fitter issue, whether it wants to use doubles in its fit (I suspect my approach would benefit from this).

pahfit/fitter.py
```diff
@@ -0,0 +1,203 @@
...
+    table. These functions have the same signature between all Fitter
+    implementations, so that the Model class can use a single
+    implementation to set up the Fitter. The Model has access to the
+    Features table and the instrument model, and needs to set up Fitter
+    with the correct initial values, bounds, and "fixed" flags (e.g.
+    setting a fixed FWHM based on the instrument for the lines). After
+    all the components have been added, the finalize_model() function
+    can be called to finish setting up the internal astropy model. After
+    this has finished, fit() can be called to apply the model and the
+    astropy fitter to the data.
```
jdtsmith, line 39: Pretty confused about this description, and why a "finalize" is needed.  It also references AstroPy.

jdtsmith: So is `Model` just "spoon-feeding" the rows of the `Features` table to a `Fitter` object?

pahfit/apfitter.py
```diff
@@ -0,0 +1,420 @@
...
+
+        """
+        self.component_types[name] = "dust_feature"
+        kwargs = self._astropy_model_kwargs(
+            name, ["amplitude", "x_0", "fwhm"], [power, wavelength, fwhm]
+        )
+        self._register_component(Drude1D, **kwargs)
+
+    def register_attenuation(self, name, tau):
+        """Register the S07 attenuation component.
```
jdtsmith, line 201: We may add other attenuation models someday, in fact have a `model` parameter in the attenuation feature just for that.

pahfit/fitter.py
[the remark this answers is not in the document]
drvdputt: In my mind `Fitter` should be told what components to fit, and with which initial values, bounds, and fixed parameters to use.  
And `Model` knows how to combine `Features` and `instrument` into the right initial values, bounds, and fixed settings. In my opinion, the way the `Features` and `instrument` are interpreted and combined, should not depend on the type of `Fitter` being used. Therefore, I abstracted out the task of adding features, so that a single logic loop can be implemented in `Model`.

> __init__: arguments: a "sanitized" Features table (with all the right FWHM, etc. set), and redshift. Do whatever setup you need to do to prepare for a fit.

If we do that, then there will be two features tables. A "normal" one and a "sanitized" one. 

Though I do see where the idea of having a "sanitized" features table is coming from. I would definitely help with dynamically enabling / disabling features depending on flags / instrument / data wavelength range, without having to re-initialize the fitter. Right now, this logic is implemented in a single place in `Model`, but it has to create a new `Fitter` on the fly. to make the feature trimming happen.

> fit: taking a (set of) Spectrum1D spectr(a/um) and or instrument keyword. On fit completion, update the Features table directly, maybe adding to its meta (e.g. for covariance).

With this, will there be edits happening in the "sanitized" features table, which then need to be copied to the original features table? More generally: how would the results be presented to the user?

I feel like we should have a single class that knows how to manage Features. Not two classes (`Model` and `Fitter`) which might have slightly different logic and some inconsistencies between the "original" and "sanitized" tables.

Alternatively, the above logic could be implemented in the Fitter ABC instead of `Model`. Also the multi-segment fitting (with multiple data ranges and instruments) will make it harder to think about this...

pahfit/apfitter.py
jdtsmith, line 1: Aha, I see we are _not_ giving the `Fitter` a `Features` table to interpret and then populate with results post-fit. We had a [discussion](https://github.com/PAHFIT/pahfit/issues/257#issuecomment-1810873394) about this that I thought led us to conclude that just handing over a (sanitized) `Features` table makes sense, rather using our own internal representation to translate model parameter inputs and fit results outputs.  Eventually `Features` may have some meta-data that informs the fit (like ties).  How will those be represented?  Another generic method?

jdtsmith (commented):
Oops hit review again.

pahfit/fitter.py
drvdputt: Yes. While it might not look like the most efficient way, I felt like this way would lead to the least assumptions. Basically, I tried avoiding doing anything "smart", so that the main focus could lie on the providing a uniform interface to the setup, fitting, and result retrieval mechanics (the things that will always be different between fitting back-ends)

I admit that it does reflect the feel of the underlying astropy: you create individual components and sum them together. But to set up each components, the right numbers need to be derived from the contents of the features table.

pahfit/apfitter.py
drvdputt: The result of that discussion was indeed that we wanted to avoid any additional internal representation. 

In the `APFitter` implementation, after all the `register_*()` calls are made,  my "internal representation" is just a list astropy model components, created directly from the `Features` table numbers that were "spoonfed". So the astropy model *IS* my internal representation. There was no need at all for a more "general" format that can be shared between the different `Fitter` subclasses. On the contrary, that would hamper the efficiency and straightforwardness.

I expect that `Fitter` subclass will store those "spoonfed" numbers in a very different format. Ideally, the numbers that come in via the `register_*()` calls are "clean" (processed by `Model`) so that you can just gather then, and then consolidate them in whatever format you need in `finalize_model()`.

Anyway, those are the thoughts that went into this design. I may be wrong about how useful this actually turns out to be.

I have thought about the parameter ties before, as well as the "multi-segment" fitting. I did not come up with a clear answer on how to implement those.
- For the ties, something like `tie_params(component_name1, param_name1, component_name2, param_name2, ratio)` might work. Maybe as an additional step during the registration process. It is not clear to me how the tie information will be included in the `Features` table, and how general the tie system would be.
- For the multi-segment fitting, I think we should write down the logic first: how do we set up the model for each segment, and then perform a joint fit of those models. No idea how achievable this will by with my system.

pahfit/fitter.py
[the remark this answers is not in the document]
jdtsmith: > Therefore, I abstracted out the task of adding features, so that a single logic loop can be implemented in `Model`.

Funnily enough I think that's where I started, and I thought you talked me out of it ;).  But handing over a `Features` table is also a simple standard... just a very highly coupled one.  I do see where you are coming from here, in fact it's what I imagined might work originally.  But it comes with some complexities.

> If we do that, then there will be two features tables. A "normal" one and a "sanitized" one.

Well, or we just update the original one so as to sanitize it.  Not much is lost, if we simply mask culled features, and update FWHM's.  But then inputting that table _again_ to another `Model` would spell trouble (as it would "re-sanitize").  So I think you have it right: leave `Features` as _pure_ as possible: just a physical model, never anything about the instrument sullying it.  

But doing this translation I guess is also OK.  It should happen _once_, when the `Fitter` is created. And again "as needed" (i.e. unless the `reuse` flag is set). 

BTW, considering this, it's almost like we do _not_ want to advertise FWHM as a line parameter at all, if it has no effect on input and is not going to be updated on output (right?): it's an instrumental artifact.   But some users will want to "override" the FWHM since they "know better".   We could say any FWHM that has non-`nan` bounds is considered a "user override" and just hand that directly to the `Fitter`.

Another issue: when the capabilities of `Fitter`s diverge, we'll need some way to indicate that (for an idea, see below).  Otherwise, `Model` will try to, for example, provide `tie`'s to both, and maybe only one can handle that.

> but it has to create a new Fitter on the fly. to make the feature trimming happen.

We can't do that, we need `Fitter` to be able to be long-lived.  But this is easy I think: always create a new `Fitter` on fit, unless `reuse=True`.  

> With this, will there be edits happening in the "sanitized" features table, which then need to be copied to the original features table? More generally: how would the results be presented to the user?

I had imagined a single table for all purposes.  What do we do here now?  Get the fit results and update the table (but only non-instrument-specific-details, i.e. no line FWHM)?  For Sara profile (#288) we'll need to translate backwards to "science pack equivalent un-convolved FWHM".  That translation better be rock-solid then.  If we allow line FWHM to "leak instrument-specific information" back to the user, that's a bit of a mixed situation, but I guess we just document that. 

> the multi-segment fitting (with multiple data ranges and instruments) will make it harder to think about this...

Yes big time.  My setup for that is (in the Fitter) to duplicate features when they impact more than one distinct spectral segment.  Now `Model` takes this over, and a `Fitter` will need to be told by `Model` "fit this [NeII] features with FWHM=0.01µm for this segment, but also fit the same feature again at FWHM=0.02µm for this other segment".  But then all `Fitter`s will need to know how to do that (or complain or something).  We could have some named `Fitter` capability set, and have a `PAHFITFitterCapability` error get thrown if you try to fit list-of-`Spectrum1D`'s to a fitter which can't do `JointFit`.

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+        ----------
+        name : str
+            Unique name. Will be used to allow retrieval of the results
+            after the fitting.
+
+        temperature : array of size 3
+            Blackbody temperature, given as [value, lower_bound,
+            upper_bound]. The bounds assume the same system as the
+            features table: if they are masked, the parameter will be
+            fixed, while +inf or -inf mean unbounded.
```
jdtsmith, line 97: We have now updated the way we store bounds.  Worth transferring that (e.g. a scalar value (no bounds) or structured array with three fields)?  We should take `nan` and `mask` off the table: if we are spoon-feeding `Fitter` don't feed it anything masked, pre-translate `nan` to +-`inf`.

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+    APFitter) need to consolidate the registered components to prepare
+    the model that they manage for fitting. After this, fit() can be
+    called to apply the model and the astropy fitter to the data. The
+    results will then be retrievable for one component at a time, by
+    passing the component name to get_result().
+
+    """
+
+    @abstractmethod
+    def clear(self):
```
jdtsmith, line 50: I'm not sure what role this serves.  If we either create a new `Fitter`, or re-use the same `Fitter`, `clear`ing one has little purpose.  Less if more for an API.

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+        """Reset model.
+
+        After reset, register_() and finalize_model() can be used again.
+
+        """
+        pass
+
+    @abstractmethod
+    def components(self):
```
jdtsmith, line 59: Similarly here.  We (`Model`) _gave_ the `Fitter` its components, so we know which ones those are.

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+        """Return list of features.
+
+        Only works after finalize_model(). Will return the names passed
+        using the register functions.
+
+        """
+        pass
+
+    @abstractmethod
+    def finalize_model(self):
```
jdtsmith, line 69: OK, I guess this is a reasonable way for `Model` to tell the `Fitter` "and that's all I've got for you".

pahfit/fitter.py
drvdputt: Good idea. I was actually having trouble with this while trying to make the tests work. Right now there is still some logic that deals with the nans (to be fair they are still quite useful for determining if something is fixed based on the bounds given).

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+        The register functions below allow adding individual features.
+        The exact implementation of how features are added, and
+        finalized in to a single fittable model, depend on the
+        underlying implementation.
+
+        """
+        pass
+
+    @abstractmethod
+    def register_starlight(self, name, temperature, tau):
```
jdtsmith, line 81: OK this is cheeky, but `register_` really doesn't do it for me.  How about `add_feature_*`?

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Parameters
+        ----------
+        name : str
+            Unique name. Will be used to allow retrieval of the results
+            after the fitting.
+
+        temperature : array of size 3
+            Blackbody temperature, given as [value, lower_bound,
+            upper_bound]. The bounds assume the same system as the
```
jdtsmith, line 95: same bounds comment here and elsewhere.

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+        """Register a dust feature.
+
+        Typically a Drude profile.
+
+        """
+        pass
+
+    @abstractmethod
+    def register_attenuation(self, name, tau):
+        """Register the S07 attenuation component.
```
jdtsmith, line 130: Not just S07.  We already have a column for `model` which can be `S07` or (eventually) others.

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+        multiplicative.
+
+        """
+        pass
+
+    @abstractmethod
+    def register_absorption(self, name, tau, wavelength, fwhm):
+        """Register an absorption feature.
+
+        Typically a Drude profile. Is multiplicative.
```
jdtsmith, line 142: Always a Drude.  We also allow `geometry`.

pahfit/fitter.py
jdtsmith: We also allow `geometry`.  For now pass and do nothing with it.

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+
+        """
+        pass
+
+    @abstractmethod
+    def evaluate_model(self, xz):
```
jdtsmith, line 148: Please just `evaluate`.  This is going to make it quite hard to do `tabulate` I guess, since it's just giving "the whole model".  Can you pass whatever wavelengths you want?  How would that work with a model which contains "duplicate" features for JointFitter types?  How does it know which of those features to use?  I strongly suspect this means we need to move this kind of logic out into `tabulate`.

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+        """Evaluate the model at the given wavelengths.
+
+        Parameters
+        ----------
+        xz : array
```
jdtsmith, line 153: `lam` and `flux` please.

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+        pass
+
+    @abstractmethod
+    def fit(self, xz, yz, uncz, maxiter=1000):
+        """Perform the fit using the framework of the subclass.
+
+        Fitter is unit agnostic, and deals with the numbers the Model
+        tells it to deal with. In practice, the input spectrum is
+        expected to be in internal units, and corrected for redshift
+        (models operate in the rest frame).
```
jdtsmith, line 171: `fit` occurs in the rest frame.  If we fit for wavelength, are we prepared to "undo" the redshift correction?

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+            made during setup. See also Fitter.components().
+
+        Returns
+        -------
+        dict : parameters according to the PAHFIT definitions. Keys are
+        the same as the function signature of the relevant register
+        function. Values are in the same format as Features, and can
+        therefore be directly filled in.
+
+        e.g. {'name': 'line0', 'power': value, 'fwhm': value, 'wavelength'}
```
jdtsmith, line 206: Missing `value` after `wavelength`.

pahfit/fitter.py
```diff
@@ -0,0 +1,209 @@
...
+            Rest frame flux in internal units.
+
+        uncz : array
+            Uncertainty on rest frame flux. Same units as yz.
+
+        """
+        pass
+
+    @abstractmethod
+    def get_result(self, feature_name):
```
jdtsmith, line 190: Why wouldn't `fit` just return all of this info in one big table (or large numpy array with the right ordering)?  Rather than a `Features` table passing back and forth, we now just do a ton of method calls.  Not much of an improvement, IMO.  What about other "results" like covariance matrix, function evals, information on convergence, etc.?

pahfit/model.py
```diff
@@ -284,9 +284,19 @@ def amp_guess(row, fwhm):
...
             loop_over_non_fixed("line", "power", lambda row: some_flux)
 
-        # set the fwhms in the features table
+        # Set the fwhms in the features table. Slightly different logic,
+        # as the fwhm for lines are masked by default
         if calc_line_fwhm:
-            loop_over_non_fixed("line", "fwhm", line_fwhm_guess, force=True)
+            for row_index in np.where(self.features["kind"] == "line")[0]:
+                row = self.features[row_index]
+                if np.ma.is_masked(row["fwhm"]):
```
jdtsmith, line 292: Does that work with a structured array?

pahfit/model.py
```diff
@@ -284,9 +284,19 @@ def amp_guess(row, fwhm):
...
+                    self.features[row_index]["fwhm"] = (
+                        line_fwhm_guess(row),
```
jdtsmith, line 294: Uh, so now we _are_ sanitizing the `Features` table, the very one the user can see and operate on.  Were we going to leave it alone and unsullied by instrument artifacts?  Why couldn't this be done right when you feed a feature to `Fitter`?

pahfit/model.py
```diff
@@ -308,7 +318,9 @@ def _convert_spec_data(spec, z):
 
         """
         if not spec.flux.unit.is_equivalent(units.intensity):
-            raise PAHFITModelError("For now, PAHFIT only supports intensity units, i.e. convertible to MJy / sr.")
+            raise PAHFITModelError(
+                "For now, PAHFIT only supports intensity units, i.e. convertible to MJy / sr."
+            )
         y = spec.flux.to(units.intensity).value
```
jdtsmith, line 324: Please lets `x` -> `lam` and `y` -> `flux` everywhere.

pahfit/model.py
```diff
@@ -387,30 +399,49 @@ def fit(
...
+    def _ingest_fit_result_to_features(self, fitter: Fitter):
+        """Copy the results from a Fitter to the features table
 
-    def info(self):
-        """Print out the last fit results."""
-        print(self.astropy_result)
+        This is a utility method, executed only at the end of fit(),
+        where the Fitter is passed after Fitter.fit() has been applied.
+        Passing a different Fitter object could be useful for
+        testing.
```
jdtsmith, line 419: How would it be helpful to use a different `Fitter` than the one you just ran `fit` on?

pahfit/model.py
```diff
@@ -387,30 +399,49 @@ def fit(
...
+
+        """
+        # iterate over the list stored in fitter, so we only get
+        # components that were set up by _set_up_fitter. Having an
+        # ENABLED/DISABLED flag for every feature would be a nice
+        # alternative (and clear for the user).
```
jdtsmith, line 425: Since we spoon-fed, we could store that list, why ask `Fitter` to do so?

jdtsmith: I do agree about a `disable` flag; I think we'll need to go there at some point.

pahfit/model.py
```diff
@@ -734,158 +758,169 @@ def tabulate(
             # any other iterable will be accepted and converted to array
             wav = np.asarray(wavelengths) * u.micron
 
-        # shift the "observed wavelength grid" to "physical wavelength grid"
-        wav /= 1 + redshift
-        flux_values = flux_function(wav.value)
+        # apply feature mask, make sub model, and set up functional
+        if feature_mask is not None:
+            features_copy = self.features.copy()
```
jdtsmith, line 763: This seems unnecessary, as indexing already creates a new `Features` view.

pahfit/model.py
```diff
@@ -734,158 +758,169 @@ def tabulate(
...
+        else:
+            features_to_use = self.features
+
+        # if nothing is in range, return early with zeros
+        if len(features_to_use) == 0:
+            return Spectrum1D(
+                spectral_axis=wav, flux=np.zeros(wav.shape) * u.dimensionless_unscaled
+            )
+
+        alt_model = Model(features_to_use)
```
jdtsmith, line 774: Oh wow we make a whole new `Model` just for a component?

pahfit/model.py
```diff
@@ -734,158 +758,169 @@ def tabulate(
...
-        the fwhm field of the table. When a new model is constructed
-        from the features table, this fwhm value will be ignored.
+        Any unit conversions and model-specific things need to happen
+        BEFORE giving them to the fitters.
+        - The instrument-derived FWHM is determined here using the
+          instrument model (the Fitter does not need to know about this
+          detail).
+        - Features outside the appropriate wavelength range should not
+          be added to the Fitter: the "trimming" is done here, using the
+          given wavelength range xz (optional).
```
jdtsmith, line 852: For JointFit's this will have to be able to handle a list of such ranges.

pahfit/model.py
```diff
@@ -734,158 +758,169 @@ def tabulate(
...
-        return PAHFITBase.model_from_param_info(param_info)
 
-    def _parse_astropy_result(self, astropy_model):
-        """Store the result of the astropy fit into the features table.
+        return is_outside & is_excludable
 
-        Every relevant value inside the astropy model, is written to the
-        right position in the features table.
+    def _set_up_fitter(
+        self, instrumentname, redshift, x=None, use_instrument_fwhm=True
```
jdtsmith, line 837: What is `x`?  Not documented.

pahfit/model.py
```diff
@@ -734,158 +758,169 @@ def tabulate(
...
+                fitter.register_dust_continuum(name, row["temperature"], row["tau"])
+
+            elif kind == "line":
+                if use_instrument_fwhm:
+                    # one caveat here: redshift. Correct way to do it:
+                    # 1. shift to observed wav; 2. evaluate fwhm at
+                    # oberved wav; 3. shift back to rest frame wav
+                    # (width in rest frame will be narrower than
+                    # observed width)
+                    w_obs = row["wavelength"]["val"] * (1.0 + redshift)
```
jdtsmith, line 887: `w` -> `lam` for all such `w`'s.

pahfit/model.py
```diff
@@ -734,158 +758,169 @@ def tabulate(
...
+                    # returned value is tuple (value, min, max). And
+                    # min/max are already masked in case of fixed value
+                    # (output of instrument.resolution is designed to be
+                    # very similar to an entry in the features table)
+                    fwhm = instrument.fwhm(instrumentname, w_obs, as_bounded=True)[
+                        0
+                    ] / (1.0 + redshift)
```
jdtsmith, line 894: You should NOT be de-redshifting here, I think.  We basically look up the FWHM as appropriate at the observed wavelength and use that in the rest frame fit.

jdtsmith:
I'm not sure exactly how to proceed, I had hoped many of your improvements would be separable from the new `Model`/`Fitter` setup, but I now see that's not the case.  Maybe the right approach is to clean up some of the names and discrepancies (e.g. now we call wavelength `w`, `x`, `xz`, `wav`), but keep the structure basically as-is.  Merge to `dev`, test for a while, the if all's OK release on master as v2.6.  BUT, keeping the understanding that the `Model` <-> `Fitter` interface will probably have to change, which will require changes in `APFitter` to suit.  

In terms of user-facing things we probably won't want to change ahead:

1. Do we really "leak" instrument information into the `Features` tables?  I say _only_ for the special case of a user having put in an unmasked FWHM ahead of time do we do this, otherwise, leave masked.  This we call "leaving `Features` unsullied".  If so, `Model` will have to keep track of the "sullied fit results" somehow separately.  Right now there are a lot of expectations of sullied `Features` (e.g. in `Model.plot`).  `plot` can just add masked `fwhm's` as needed at the last moment, in fact _should_ do so. 
3. Are we certain we want users to `mymodel.features.loc['h2_s01']['power']` vs. just `mymodel.loc['h2_s01']['power']`.  I.e. isn't `Model` _mostly_ just a `Features` table with some functionality for setting up a `Fitter`?  We could easily keep the same system and provide indexing pass-through wrappers on `Model`.  So maybe that doesn't have to be decided now.

pahfit/fitter.py
jdtsmith: That's for `Model` in our new way of thinking.

pahfit/apfitter.py
jdtsmith: I think it's fairly reasonable, just not what I had envisioned.  It's basically substituting a large suite of methods for a data structure of specified type (which, you know, `Features`).  If we have a `reuse` flag available, I think this will be OK, as long as there is room for querying capability and "spoon-feeding" more interesting flavors, like "here's a feature that needs to be duplicated and jointly fitted, by segment".   The `tabulate` methodology feels really kludgy, which might indicate we need a richer interface into `Fitter` to "calculate these features at these wavelengths".  

For `tie` I have some notes in which I just add `tied` and `ratio` to `meta`.  But `Model` could be made aware of these and indeed pass via `tie_params`.  Though my tie setup is a bit more complex, since you can tie a _group_ of features to another _group_.  But conceptually, yes. 

Let's make multi/joint-fit in another issue.

pahfit/fitter.py
jdtsmith: I think this could be workable, as long as we can re-use a `Fitter` when requested.

pahfit/fitter.py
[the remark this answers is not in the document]
jdtsmith: Opened a discussion on MultiSegment/JointFitter #291  and also I do indeed now think we should "solve our problems" with line FWHM by masking it entirely.  It becomes not a settable or retrievable for the model at all.  Instead we can offer a `sigma_v` for some "additional" velocity spread.  We might leave open a hidden workaround for our own uses.  See #293.

pahfit/apfitter.py
jdtsmith: Yes, those are there waiting.  So you propose to leave those ignored in the API for now?  I suppose that's OK.  We will certainly have the problem when only one of the `Fitter`s implements something.  I suppose we can update the other to accept the argument and give an error or something.  Or we need a "Capabilities" flag setup that can be queried (`SupportAltAttenuationModels`, `SupportAttGeometry` etc.).  BTW, mixed vs. screen is in the original PAHFIT model, so should probably be supported OOTB.

pahfit/apfitter.py
jdtsmith: See #291  for multi-segment.  I do like the idea of a final pass of method calls for ties.  See #291 for an idea on segments. These are both TBD, but should be workable.  I'm coming around to your simple system.

pahfit/fitter.py
jdtsmith: We can indicate fixed by passing a scalar value in the spoon-feeding stage.

pahfit/fitter.py
jdtsmith: E.g. could run after `tie` if/when that materializes.

pahfit/fitter.py
jdtsmith: Philosophical question: can a `Fitter`, the _same_ `Fitter` you earlier fit data with, be used to `evaluate_model` on a different set of wavelengths?  It has all its own internal ideas of `FWHM` for lines and everything, so in principle it could.  But what about multiple segments, with multiple values for `FWHM` for the same line?  Maybe the rule is: must use same exact bounds (can increase point density if you want), and same shape e.g. (list of N segments).  Enforced, or convention?

pahfit/model.py
jdtsmith: I think this is because you've been thinking in terms of building a new `Fitter` each time.  For testing I'd say make a new `Model` with a different `Fitter`, and pre-seed it with the `Features` table from the old model.

pahfit/features/features.py
drvdputt: Python floats are implemented as `double` in the Python source code (which is C). So the astropy modeling also uses doubles. I made the above change because the results were actually subtantially different. I think it was my test for writing and reading a `Model` object. It checks `np.testing.assert_allclose(column1, column2, rtol=1e-6, atol=1e-6)`, there were small differences (even with only a 1e-6 tolerance). Not sure why writing/reading would result in these changes, but changing the data type worked.

jdtsmith: If it writes to ASCII, the number of digits allocated sets the precision.  I know that scipy can use either float or double and will respect your data type.  Can look into it.

pahfit/fitter.py
drvdputt: Yes we should definitely deal with these details in `Model`. I will try to make it work with the scalar vs 3-array approach. That does feel less cryptic, so I made the change and pushed it. I implemented the array/scalar check using `__hasattr__(array_or_scalar, '__len__')`. Seems to run safely according to the tests.

pahfit/fitter.py
drvdputt: I'm not using it anywhere. Can't remember what my intended use was, so will remove it again.

pahfit/fitter.py
drvdputt: I feel like this might come in handy. I'll come back to this once I've cleaned up the rest.

If it turns out that we need to remember which components we gave it, then 
- either store it as part of the `Fitter` (fitter says "how was I set up")
- or we need to store an extra value in `Model` ("how did we set up Fitter? And maybe with the re-usable Fitter: which of its components are active?") 

(On a side note, a dictionary can provide a list of its keys, even though you "should" already know what those are because you gave them to it)

pahfit/fitter.py
jdtsmith: Inside your `Fitter` you mean?  Are you passing the val/min/max flavor when there are bounds?

pahfit/fitter.py
drvdputt: Sounds good to me.

In the `APFitter` implementation, I'm also changing the wording. I use 'component' when using astropy lingo, i.e. everything related to setting up the multi component model (so only for hidden functions, variable names, and implementation comments).  But "feature" for the PAHFIT concepts.

pahfit/fitter.py
drvdputt: The `Fitter` setup logic (implemented in `Model._set_up_fitter()` cleans the values before they are passed to `add_feature_*()`, and will pass a scalar or a 3-array as appropriate. The above check happens inside `APFitter`, where it checks if it receive a scalar or not.

pahfit/fitter.py
drvdputt: Addressed by a9311cc

pahfit/fitter.py
drvdputt: Adding this to indicate our intentions.

pahfit/fitter.py
drvdputt: placeholder `geometry` argument also added here

pahfit/fitter.py
drvdputt: I do not have a clear answer for this.

For multiple segments, I speculate that there would be multiple model functions, each one to be compared to a certain segment during the fitting (and of which the chi2 are summed together). In that case, and index could be passed, to indicate which part of the "multi-function" should be evaluated.

I think we should first work out how we can reuse the `Fitter` to tabulate individual or groups of components. Perhaps passing a list of feature names to `evaluate()` would be an option? In that case, the evaluation can be generic (implemented in the ABC `Fitter` using the generic component evaluation functions) or specific (making use of the individual astropy components) as necessary.

If we want to avoid duplicating any logic, having it in `Model.tabulate` somehow, or the `Fitter` ABC, would be best.

jdtsmith: This I think is an argument in favor of implementing the standard set of `Fitter.gaussian/drude/blackbody` etc. components.  There are really only a few of those.  Then we could do away with `Fitter.evaluate` altogether.  Evaluation is not (really) the `Fitter`'s job.  The fitter's job is to do the fit internally using whatever methods and data structures it wants, and return to us the results.  

Beyond that, for `Model`'s convenience, we'd also like each `Fitter` to provide access to the _component functions_ with specified arguments (units, order, etc.) and output units.  Maybe for simplicity and convenience we can ask it for a full `Fitter.attenuation_function`, providing to that method `geometry`, `model`, `tau_si`, and details for other attenuation `tau`'s, by which we can multiply our spectra, or components of our spectra (TBD: increase the `tau` FWHMs for LSF: Sara Profile again).

With this in place, `tabulate` is all we need, and it can be generic.  It just looks at the (set of) spectra it's being called with, does the same parameter marrying with the instrument pack that it does for `Fitter.fit`, and sums up the various `fitter.drude(x,y,z)`, etc., finally multiplying by the `attenuation_function`.

Yes, this is additional work that is broadly similar to what each Fitter is doing internally, but on another level it serves as a good "output comprehension" test.  If you use the `Fitter`'s results to tabulate the model and it doesn't actually match the spectrum, you have not understood the output.

We can even add a test to demand that all `Fitter`'s, when given some standard input, produce the ~same output for each of the component types.

pahfit/fitter.py
drvdputt: Did a find and replace. Some cases where there was 'x' and 'y' too, I've called them lam_obs, and flux_obs (in the observer frame), to distinguish them from lam and flux which are in the rest frame..

pahfit/fitter.py
drvdputt: For me, fitting the wavelength makes the most sense in two cases
1. a small correction to the rest frame wavelength of a line (to deal with e.g. small wavelength calibration issues)
2. determining the rest frame wavelength of a dust feature (which can shift for some of them, depending on the physics)

These cases have not much to do with redshift (other than the fact that we need to shift the observed data to the rest frame, to make such fitting work). When there are actual velocities causing the lines to shift, things are a bit more abiguous. 

But no, at the moment there is no utility to convert the rest frame values back to the observer frame. It might be useful to implement this somewhere, as a utility for visualizations (direct comparison to the data).

pahfit/fitter.py
drvdputt: With the astropy framework, I can't "just" return a big table. I have to loop over the components and gather the results. 

So in my mind, every subclass have to implement some logic to create this table, and we would have to decide on a common format and guarantee that it is consistent between `Fitter` subclasses. (this might be another argument for just passing a Features table directly to the `Fitter`, and implementing the Features <-> fitting/results logic in the ABC).

As it currently stands `Fitter` only offers a way to retrieve a "single unit" of results. And the logic for how to map those units to the features table, is implemented in a single place: `Model._ingest_fit_result_to_features()`.

For the covariance matrix, we will have to come up with something else... but I feel it will still be possible to try and have a single place for the logic.

The fitting diagnostics / convergence information can be wildly different depending on the underlying fitting algorithm. I feel like understanding them full, will always require some knowledge of the specific framework. Though, if we can come up with a few "standard" diagnostics, then we can just provide access functions for those.

pahfit/fitter.py
jdtsmith: Sorry I didn't word that well.  I meant "correcting" for redshift is something we ourselves do in `Model`, yes?  I.e. the user doesn't have to `lam_rest = lam/(1+z)` which this doc makes it sound like.  Do we accept a `z` on a model and/or in a `Spectrum1D`?

pahfit/model.py
drvdputt: As far as I know, the element (features[row_index][col_name]) can still be masked in its entirety. I.e., it can be equal to np.ma.masked, or it can be a structured array element (the one we defined with 'val', 'min', and 'max')

pahfit/fitter.py
jdtsmith: I've definitely come around to your way of thinking on this since that old comment.  For "fit information", we can just ask the Fitter for some generic fit information for `Features.meta["fit_log"]` and insert it,  perhaps a dictionary, so each Fitter can define their own (with some attempt to normalize things e.g. `Features.meta["fit_log"]["function_evals"]` etc.). 

We can (later) ask the `Fitter` for the `covariance` too, and store it in `meta` as well.  We can plan to access that in a `Model.total_power` (i.e. total up various powers of the selected table, with uncertainty propagated through the full covariance matrix). And of course for `Features` pretty-printing.

pahfit/model.py
drvdputt: Leaving it alone would definitely be best. I think it was a bit of a hack to let the `guess()` step set the fwhm. I will give it some thought, and see if I can make the main use cases work without guessing `fwhm` for the lines in this step. Then we just need to make sure that the right fwhm values are generated on the fly, e.g. when plotting etc, but also that we don't accidentally overwrite any fwhm values that were fit.

pahfit/model.py
drvdputt: I have done `x` -> `lam_obs`, `y` -> `flux_obs`
To separate it from
`xz` -> `lam`, `yz` -> `flux`

The former is observer frame, the later is rest frame.

pahfit/model.py
jdtsmith: I agree: leave it alone.  If `Features` is ever and always fully and completely divorced from any aspect of instrument information, `plot` will need to "just in time" marry the instrument information from the spectrum(/a) it is plotting.  As will `fit` (modulo "re-use") and `tabulate`.  And (in another way) `guess`.  But all of them get it "just in time" (for convenience we can calculate and cache it perhaps).  `Model` probably needs a function to do this.

For lines, the simplest solution, and one I'm advocating now, is to treat line FWHM as a fixed **instrument exclusive**.  None of your business.  See #293.  `guess` will need to "know about it", since it has to guess power from a flux, but we have that information from the spectra being "guessed".  Line FWHM never appears in the `Features` table.  Setting it on lines there does nothing, just as it does nothing for starlight.  

The real advantage of this _pure science_ `Features` approach is that _any_ spectra or combination of spectra can be used to update `Features`.  If you choose to plot or re-fit it using a very different flavor of spectra, so be it, you probably have your reasons.  We will "do the right thing" with the instrument info you gave us.

As for stitched spectra, `instrument` already provides an intermediate resolution, which should be good enough for most types of stitching.  Users can always _cheat_ by opening up `sigma_v` on lines, but they will then know they are cheating.

I recognize this could be a significant change, so for now let's stick to the semi-hack line and guess FWHM approach we have.  Then after v2.6 is out and the other fitter is underway, we can update the API, move to remove it, add `sigma_v`, and update docs (for v3).

pahfit/fitter.py
jdtsmith: Sorry I was confused about that.

pahfit/fitter.py
drvdputt: Fixed.

pahfit/model.py
jdtsmith: Thanks.  See also #294; I believe we have the "shift to rest" wrong for the flux.

pahfit/model.py
drvdputt: Ok, I added a TODO comment in guess(). We can address this after the merge into dev via #293

pahfit/model.py
jdtsmith: How would a structure array have just a single value, though?  I thought our method for "fully masking" a given parameter is to mask the value?

pahfit/model.py
drvdputt: I used this in a test, but I changed the way things work now. Instead of returning / taking a Fitter object, the internal utility functions of `Model` now just set and edit `self.fitter`

pahfit/model.py
drvdputt: E.g. 
```
In [12]: f.loc['PAH_6.2']['wavelength']
Out[12]: (6.22, nan, nan)

# "starlight" does not have the "wavelength" parameter -> masked
In [9]: f.loc['starlight']['wavelength']
Out[9]: masked

In [14]: np.ma.is_masked(f.loc['starlight']['wavelength'])
Out[14]: True
```

pahfit/model.py
drvdputt: I now store this list as `self.enabled_features`.

pahfit/model.py
jdtsmith: I think that's just for display.  Underlying that is still a structure array member.  I believe masking _any_ of the structure array members results in this.

pahfit/fitter.py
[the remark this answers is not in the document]
jdtsmith: I think we've converged here (mostly on your plan :). TBD is whether `Model` knows how to create a model from the `Features` table itself, only asking `Fitter` for the components.

pahfit/model.py
drvdputt: It can be any combination of components that needs to be tabulated, not necessarily a single component. (Also, remember that fact that `tabulate` was born out of the format `sub_model`). 

This can be changed once we have our standardized component evaluation routines.

pahfit/model.py
```diff
@@ -313,23 +316,23 @@ def _convert_spec_data(spec, z):
...
 
         # transform observed wavelength to "physical" wavelength
-        xz = x / (1 + z)  # wavelength shorter
-        yz = y * (1 + z)  # energy higher
-        uncz = unc * (1 + z)  # uncertainty scales with flux
-        return x, y, unc, xz, yz, uncz
+        lam = lam_obs / (1 + z)  # wavelength shorter
+        flux = flux_obs * (1 + z)  # energy higher
+        unc = unc_obs * (1 + z)  # uncertainty scales with flux
+        return lam_obs, flux_obs, unc_obs, lam, flux, unc
```
jdtsmith, line 335: See #294.

pahfit/model.py
drvdputt: More precisely, basic slicing returns a view, while advanced indexing returns a copy. I did a copy() as a safeguard against unwanted modifications. Might not be necessary after all. Removing copy for now.

pahfit/model.py
drvdputt: Changed to `lam` and added more documentation.

pahfit/model.py
drvdputt: Note that to do the rest frame fit, we are using modified data, as we correct for the redshift before the fit. As the wavelengths are shortened, the line profiles as seen in the data will become narrower. This change should be reflected in our model, by making the fwhm narrower by the same factor.

drvdputt:
> Lines 183-186 of model.py:
> 
> ```
>         inst, z = self._parse_instrument_and_redshift(spec, redshift)
>         # save these as part of the model (will be written to disk too)
>         self.features.meta["redshift"] = inst
>         self.features.meta["instrument"] = z
> ```
> 
> Is this correct? It seems that these definitions for redshift and instrument are flipped?

This has now been fixed (see latest commits pushed).

I have also gone through most of the other comments. So we should be close to merging this, and then starting a new pull request to integrate my implementation of the power-features.

A few things revealed by the tests:
- @jdtsmith Linkcheck says the link to your website is broken (this one: http://tir.astro.utoledo.edu/jdsmith/research/pahfit.php)
- Codecov is complaining about needing a token. Not super important right now. I think we should fix all our continous integration services in a single pull request. Will take another look at the reports and summarize in an issue.
- I did a second push to fix a double space somewhere.

pahfit/apfitter.py
drvdputt: I created pahfit/fitters (with an `__init__.py`), and did the following moves
pahfit.fitter -> pahfit.fitters.fitter
pahfit.apfitter -> pahfit.fitters.ap_fitter
pahfit.component_models -> pahfit.fitters.ap_components

pahfit/model.py
jdtsmith: Good point.  Redshift is annoying.  The model itself needs to shift to the rest frame, and for instrument details, your narrow line got narrower (fractional FWHM is a better parameter for that reason, since it doesn't change with redshift, but that ship has sailed).

For dust features, we specify their positions and widths _in the rest-frame_, so we simply evaluate them there with those (our de-redshifting has de-expanded them).  

When we expand Drude profile widths ala #288, we'll need to do something similar with the teleported line profile width, i.e. providing "the width of the line profile at the observed center of the feature, cast into the rest frame".  We should probably clearly spell this out in a comment here.

pahfit/fitter.py
drvdputt: The list of "active" features is now stored in `Model` as `self.enabled_features`

jdtsmith:
Thanks for getting back into it Dries.  My server has been cutoff at the firewall, working to restore access.  I agree about the CI.  Do we need to solve #294 now too given all the redshift discussion?

pahfit/model.py
drvdputt: I ran into a bug due to this. Apparently, in our current setup, `row['fwhm']` can be either 0-dimensional or 1-dimensional. (with `row` a row of the features table)

This is a problem, because if we want to check if 'fwhm' is masked, then `row['fwhm'][0]` will yield an error if `row['fwhm']` is 0-dimensional. Vice versa, in the 1-dimensional case, we also get an error if we try to use `np.ma.is_masked(row['fwhm'])`.

A workaround would be checking `row['fwhm'].ndim` each time.

drvdputt: To be clear, if the above displays `masked` then we have the 0-dimensional case. So yes, it really is a single value somehow. Even more striking: if I check the data attribute, i get this
```
ipdb>  p row['fwhm'].data

array(0.)
```

jdtsmith:
Update: my server will unfortunately not be allowed back on the external internet (new VPN-only policy); I've uploaded "PAHFIT classic" to https://github.com/PAHFIT/pahfit_classic.  Can you update the internal links to point there? 

Aside from the redshift discussion which seems to have wound down, what else is needed on this PR?

pahfit/model.py
drvdputt: This has been rephrased as "we do the necessary adjustment as follows"

drvdputt:
In addition to your website, there is an additional documentation issue  I discovered. So I just pushed two new link fixes.

The biggest issue still remaining is cases like this:
```python
            for row_index in np.where(self.features["kind"] == "line")[0]:
                row = self.features[row_index]
                if np.ma.is_masked(row["fwhm"]):
```

As in the comments above, row["fwhm"] can be a single element, or a tuple (so 0D or 1D). I ran into this issue while running one of my demo notebooks. There is no "readable" check that works for both. But I have a hacky workaround, which replaces some of of the `np.ma.is_masked` checks with `.ndim == 0`.

jdtsmith:
>  But I have a hacky workaround, which replaces some of of the `np.ma.is_masked` checks with `.ndim == 0`.

Did you see my `row["fwhm"] is ma.masked` method above?  That seems to be clearest.

drvdputt:
Looks promising doing some tests in the python shell. I will give it a shot and see if I can get my notebook to run.

drvdputt:
Seems to work fine for now. But at one point, this logic will have to be revised, when we set up a clear way to allow users to fit the FWHM of specific lines. Will be addressed by #293.

Let's leave the redshift issue as is for now, so we can finish the work on the power features too. Once this has been merged, I can rebase my power features branch, and spin up the next pull request.

jdtsmith:
OK, great, shall we go ahead and merge this to `dev`?

drvdputt:
I think this is good to go. I will be actively using the dev branch (with my power features applied), so I will see if anything pops up along the way.

Issue #299, opened 2024-10-14
Alternative dust continua


drvdputt:
Problem: the standard Modified blackbody -- Blackbody(lambda; T) * (lambda**-2) -- shape for the dust continuum does not work for certain objects.

Idea: the lambda**-2 term represents an absorption coefficient of the dust. As a start, we could replace it by an empirical extinction curve. This is already in the works in the PAHFIT branch for a paper I'm writing.

I also want to highlight the shape of the continuum of the CAFE tool, shown in e.g. this paper https://arxiv.org/abs/2410.09020 (I'm writing this up now because this recent paper reminded me of it). The hot component has a silicate emission bump at 10 micron, which is partially reduced again by the silicate bump in their dust attenuation model. So there is a precedent for such a continuum shape being reasonable.

Implementation: To switch between the "default" MBB and alternative ones, we can use the "model" column that is already in the Features table. The backends could then to implement the available models, and provide an argument in the `add_feature_dust_continuum()` function of the `Fitter` interface.

drvdputt:
I just realize this is a duplicate of #11, where some more info is given no what the problems are

jdtsmith:
Historically I've shied away from silicate emission in the models, since emission and absorption tend to "fight" and you enter a regime where radiative transfer becomes the more appropriate tool.  CAFE's continuum is interesting, but I know from users that it can be highly sensitive to starting point and input choice.  Perhaps we could experiment with a few models for the opacity function for an MBB.

karllark:
Definitely an issue with silicate emission/absorption fighting with each other.  On the other had, silicate emission is definitely real, hence we should definitely have the option to include different emissivity relationships other than the simple powerlaw.   Especially as the powerlaw is not like any dust emissivitiy law I'm familiar with for the NIR/MIR.

jdtsmith:
Good thoughts.  One idea to avoid the fight that's working well from M51 is to use H-recomb to form an independent estimate of tau_Si, fix that (or with a small range).  Then you could allow a more complicated MBB emissivity to sit "on top" (or behind/mixed).  We may need a bit more expressiveness in the science packs to accomplish this.

## Absorption/Attenuation

- "Normal" attenuation curve (power-law + Si-O resonances), perhaps with a few sub-options, e.g. one with high crystalline fraction.  Karl, we should discuss the current best "updated" curve to include.  I know the 18µm resonance is challenging for full self-consistent UV-IR.
[...]

Issue #298, opened 2024-09-25
Modular science pack idea


drvdputt:
In my recent work, I have started a new habit where I treat science packs in a modular way.

An example use cases is as follows:

You have two objects:
1. A PDR
2. A galaxy

And you would like a science pack situation as follows
- dust_feature tuning: keep it the same for both. And you are making frequent adjustment as testing them on both objects.
- dust_continuum tuning: custom for each. And the galaxy needs a starlight component too.
- (lines: separate pack for PDR and galaxy)

In that case you could put the dust features in a general file, and the continuum in separate ones
`dust_features.yaml, continuum_pdr.yaml, continuum_galaxy.yaml`

To load them and combine them into a single science pack, I wrote the following function
```python
from astropy.table import vstack
def make_multi_yaml_model(*args):
    return Model(vstack([Features.read(fn) for fn in args]))
```

The minimum effort to implement, would be to just add allow `Model` to take a list of YAML files at setup, and make it apply this function. In practice, it would probably be best to include a few checks though. Most importantly, making sure that all feature names are unique.

jdtsmith:
This is interesting.  Separate line packs might also be very smart, since non-existent lines chew up CPU, and sometimes you just don't need that many.  A separate continuum pack makes sense too.   

In terms of what packs to distribute, this is harder.  We don't want to make it easy for users to shoot themselves in the foot, so shipping "blessed combinations" make sense.  From that point of view, extending our YAML `Features` syntax to allow referencing other sibling YAML files might make the most sense, e.g. an entire `pdr.yaml` pack could read:

```yaml
scipack_include: pdr_lines
scipack_include: classic_continuum
scipack_include: classic_starlight
scipack_include: classic_attenuation
scipack_include: ice_absorption
```

This would include `pdr_lines.yaml`, etc. (and warn about any duplicate names).   Of course you can directly remove features from the `model.features` table itself.

One other issue is tied constraints, which may bridge modular input packs.  But we can cross that budget when we get to it.

Issue #296, opened 2024-07-09
Units of (modified) Blackbody amplitude


drvdputt:
Currently, `tau` is used as the strength of the blackbody (stellar continuum) and modified Blackbody (dust continuum) components, and the equation for evaluating the blackbody looks like this

```
            amplitude
            * 3.97289e13
            / x**3
            / (np.exp(1.4387752e4 / x / temperature) - 1.0)
```

where the fitted amplitude is reported as `tau` in the `Features` table.

In general, the fact the we often have to deal with the tails of the blackbody profile, makes them hard to use. In my experience, the typical order of magnitude of the resulting `tau` values can be very different, scaling strongly with the temperature of the blackbody. I feel like this causes numerical issues sometimes; very often, the minimizer will use only two components, while all the rest of the tau values are set to 0.

We could think of an alternate parameterization / normalization for these components. A few ideas:
- Amplitude at the peak wavelength. This becomes a problem however when the peak is far "off screen", and only the tail of the blackbody is seen
- Amplitude at a fixed wavelength, say 20 or 30 microns.  Would be ok for most dust components, since at most dusty temperatures, the blackbody profile will have a reasonable contribution there. This will not work for the stellar component however.
- Total power: we could introduce something like "PowerBlackbody". Again, this might be problematic for components that peak far outside the wavelength range. But at least it would allow us to set reasonable and perhaps physically motivated upper limits.

jdtsmith:
Missed this.  

I'm not actually sure the zero amplitudes result from numerical instability or "tail fitting", but rather their unimportance to the final continuum shape.  $\tau$ _is_ roughly the "amplitude at the peak", which as you say may be far higher and off to the red from the fitted wavelength regime.  I don't think total power of a blackbody makes a lot of sense, since that is not anywhere constrained, and might encourage people to interpret it as total IR luminosity.  Fixed wavelength amplitude might be OK, but given the fixed temperatures (typically) this would just be a constant numerical scaling of the amplitude by some set of factors of order 1-10 (guessing).  

$\tau$ does intrinsically have surface brightness unit scaling.   $\tau_\star$ can thus be interpreted as the fraction of the aperture occupied by the surface of stars (which we consider equivalent).  We could normalize that to some fixed fractional value and use those for our internal $\tau$ units, but I'd like to be convinced there's a numerical issue in the first place.

jdtsmith:
> very often, the minimizer will use only two components, while all the rest of the tau values are set to 0.

[...]

Issue #149, opened 2021-09-22
No uncertainties being returned with LevMar fitter


karllark:
As noted in PRs #141, #142, #143, we are not getting uncertainties when using the LevMar fitter interface provided by in astropy.modeling.  Initial thoughts were that the problem could be the Compound astropy.model we use or that some of the components in this Compound model were not "stock" astropy functions.

Work by @alexmaragko showed that it was not the Compound model.

Continuing this investigation, I have been able to show that it is not the use of non-"stock" models.  The issue seems to be when a parameter is not needed or is at one of the limits (zero in the example 2 below).  This is not an issue with astropy.modeling, but is the result of the underlying scipy function not returning a covariance matrix.  Astropy.modeling uses the scipy returned covariance matrix to calculate the uncertainties.

1st example works - uncertainties reported.   2nd example does not report uncertainties.  Only difference is that the simulation  has tau_sil  = 0 instead of tau_sil = 1.

Example 1 of fitting a Compound model with (Drude + line) * S07_attenuation model (see code below).  The tau_sil = 1 in the simulation.

![Screenshot from 2021-09-22 15-56-30](https://user-images.githubusercontent.com/1554536/134415820-1a822baf-aa6e-4487-a0a5-4453ef70960b.png)

> fit_info message
> Both actual and predicted relative reductions in the sum of squares are at most 0.000000
>
> fit_info param_cov from scipy
> [[ 7.71439959e-03 -1.13153353e-04 -3.08345771e-03 -9.72135409e-06 8.85727485e-04  5.37435623e-03]
>  [-1.13153353e-04  3.95494604e-04 -5.27363936e-05 -3.62103846e-06 -5.54130404e-06 -3.62487661e-04]
>  [-3.08345771e-03 -5.27363936e-05  5.54494824e-03  1.72732595e-04 -3.66794776e-03 -3.67701879e-03]
>  [-9.72135409e-06 -3.62103846e-06  1.72732595e-04  2.21930261e-05 -3.87891140e-04 -4.76215557e-04]
>  [ 8.85727485e-04 -5.54130404e-06 -3.66794776e-03 -3.87891140e-04 7.77305065e-03  1.17720169e-02]
>  [ 5.37435623e-03 -3.62487661e-04 -3.67701879e-03 -4.76215557e-04  1.17720169e-02  3.13437185e-02]]
>
> astropy modeling uncertainties
> standard deviations
> amplitude_0| 0.088
> x_0_0      | 0.02
[...]

PR #308, opened 2025-09-11
Porting features from the Orion Bar paper branch

drvdputt:
There are still some loose ends since the publication of my paper (https://ui.adsabs.harvard.edu/abs/2025arXiv250705848V/abstract)

1. There are some additional features only available in my fork of PAHFIT. 
2. We have to add the PDR pack. I have made a clean version of it.

For now, to ensure reproducibility, I have created a snapshot of the code as it was used for the Orion Bar paper, in the form of a release under my fork: https://github.com/drvdputt/pahfit/releases/tag/OrionPaper2025.

I am submitting my branch as a pull request, so you can see the changes. We can decide which features we want to port over and clean it up, though it looks like it's not that much in terms of lines of code. 

The most notable parts, with some relevant issues linked:
1. Option to switch between dust continuum models, by providing the "model" keyword in the PDR pack. + The modified blackbody that I used in the paper, here called `SpecialModifiedBlackbody1D` #299 
2. Option to switch between fitting methods (different algorithms can be selected within APFitter). Currently this is done by passing down a keyword from the `Model.fit()` invocation. Options are `"lm"` and `"trf"` (Levenberg Marquardt vs Trust-Region Reflective). The `Fitter` subclass has a function to advertise the available options. Relevant for #149 
3. Different normalization for the blackbodies, addressing numerical precision issues. As discussed previously, we can consider changing the units of `tau` instead. (also: adjust initial guess for `tau`, as it was clipped between 0 and 1 previously). #296 
4. Option to pass multiple science packs when initializing a `Model`. Backwards compatible with a single science pack by using `*` for a variable number of arguments. #298

### DIFF
--- a/pahfit/features/features.py
+++ b/pahfit/features/features.py
@@ -32,7 +32,7 @@ import pahfit.units
 
 # Feature kinds and associated parameters
 KIND_PARAMS = {'starlight': {'temperature', 'tau'},
-               'dust_continuum': {'temperature', 'tau'},
+               'dust_continuum': {'model', 'temperature', 'tau'},
                'line': {'wavelength', 'power'},  # 'fwhm', Instrument Pack detail!
                'dust_feature': {'wavelength', 'fwhm', 'power'},
                'attenuation': {'model', 'tau', 'geometry'},
--- a/pahfit/fitters/ap_components.py
+++ b/pahfit/fitters/ap_components.py
@@ -41,7 +41,9 @@ class ModifiedBlackBody1D(BlackBody1D):
 
     @staticmethod
     def evaluate(x, amplitude, temperature):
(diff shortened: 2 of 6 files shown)


jdtsmith:
Awesome, thanks for your work, and for the reminder to take a look.  Apologies for the long delay.  I think we should pull this apart into a few PRs.  Maybe something like (starting from highest priority):

1. `astropy.modeling` "flavor (`TRF`) passthrough.
3. PDR pack (does this use the alternative continuum flavor?).
4. Pack inheritance.

I'm a bit worried about the special modified blackbody, that seems to open a real can of worms.  One idea is to maintain that as a separate repo (with appropriate warnings).

karllark:
I respectfully disagree about the special modified blackbody.  I think this is an important enhancement.

PR #310, opened 2026-01-14
improved plot function. used Leg_lines colours and marker sizes direc…

vanshaj18:
used Leg_lines colours and marker sizes from the Leg_line list. Added vertical separation between plot and residuals for visual separation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ develop-eggs
 distribute-*.tar.gz
 
 # Other
+.env
 .cache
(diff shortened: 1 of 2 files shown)


jdtsmith:
Thanks for the PR.  We are making a significant push on PAHFIT this summer, so may ask you to retarget your PR to the dev branch (once it settles).  Since we have the capability of multiple "fitting cores" we'd probably need to make the statistics reporting follow that API.

PR #313, opened 2026-06-22
Add covariance uncertainty output helpers

hcha9:
Summary:
- Adds helper output for covariance-derived parameter uncertainties.
- Adds a feature-table display helper with numeric uncertainty columns and readable value \pm (plus minus) uncertainty columns.
- Keeps the full covariance matrix available as an optional diagnostic.
- Mostly distilled and preserves PAHFIT Ver.2.6.99
        - Preserves access to both LM and TRFLSQ fitting through the existing fitter method interface.
        - Preserves bounded/frozen parameter handling when reporting fitted values and uncertainties.
 
Notes:
- The reported uncertainties are formal covariance-based uncertainties from the Astropy fitter (fit-statistical uncertainties).

### DIFF
--- a/pahfit/error_functions.py
+++ b/pahfit/error_functions.py
@@ -1,0 +1,251 @@
+#After running PAHFIT model.fit you can use 'model' as the input to pahfit_to_ecsv to save all fit params with associated errors from covariance 
+#matrices. The rest of the functions in here are just helper functions to make pahfit_to_ecsv work. -Hannah
+
+import numpy as np
+from astropy.table import Table
+
+
+def build_flat_name_map(astropy_model):
+    return {v: k for k, v in astropy_model._param_map.items()}
+def get_covariance(model, comp1, par1, comp2, par2):
+    astropy_model = model.fitter.model
+
+    flat_name_map = build_flat_name_map(astropy_model)
+
+    # component index
+    idx1 = astropy_model.submodel_names.index(comp1)
+    idx2 = astropy_model.submodel_names.index(comp2)
+
+    # map to flat names
+    flat1 = flat_name_map[(idx1, par1)]
+    flat2 = flat_name_map[(idx2, par2)]
+
+    return astropy_model.cov_matrix[flat1, flat2]
+def get_power_and_error(model, feature_name):
+    astropy_model = model.fitter.model
+
+    comp = astropy_model[feature_name]
+    power_val = comp.power.value
+
+    try:
+        var = get_covariance(model, feature_name, "power",
+                                   feature_name, "power")
+        power_err = np.sqrt(var)
+    except KeyError:
+        power_err = None
+
+    return {"power": power_val, "power_err": power_err}
+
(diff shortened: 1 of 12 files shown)


pahfit/error_functions.py
```diff
@@ -0,0 +1,251 @@
+#After running PAHFIT model.fit you can use 'model' as the input to pahfit_to_ecsv to save all fit params with associated errors from covariance 
+#matrices. The rest of the functions in here are just helper functions to make pahfit_to_ecsv work. -Hannah
+
+import numpy as np
+from astropy.table import Table
+
+
+def build_flat_name_map(astropy_model):
```
hcha9, line 8: Converts Astropy's internal flattened param map into a lookup table (Astropy compound models store covar. by flattened param names).

pahfit/error_functions.py
```diff
@@ -0,0 +1,251 @@
...
+    return {v: k for k, v in astropy_model._param_map.items()}
+def get_covariance(model, comp1, par1, comp2, par2):
```
hcha9, line 10: Fetches one covariance-matrix element.
e.g., get_covariance(model, "[NeII]", "power", "[NeII]", "power")
This shows the variance of [NeII] power.  If two different params used, it returns the covar. between them.

pahfit/error_functions.py
```diff
@@ -0,0 +1,251 @@
...
+    # component index
+    idx1 = astropy_model.submodel_names.index(comp1)
+    idx2 = astropy_model.submodel_names.index(comp2)
+
+    # map to flat names
+    flat1 = flat_name_map[(idx1, par1)]
+    flat2 = flat_name_map[(idx2, par2)]
+
+    return astropy_model.cov_matrix[flat1, flat2]
+def get_power_and_error(model, feature_name):
```
hcha9, line 24: Convenience helper for one common case: get power and power_err

pahfit/error_functions.py
```diff
@@ -0,0 +1,251 @@
...
+        if verbose:
+            print(f"Saved to {output_file}")
+
+    return table
+
+
+
+
+
+def covariance_to_table(model):
```
hcha9, line 171: This is diagnostic only. For normal science output, use features_with_uncertainties() instead.

drvdputt:
Some interesting features here. Can you point the pull request to the dev branch instead of master? That should make it more clear what is new; right now, many other commits seem to be included.

It looks like some discussion about code organization will be necessary, since getting the errors this way is backend-specific (only works with ap_fitter).

Ideally, this uncertainty code would be built upon a version of PAHFIT with a cleaned up version of #308 pulled in. But we could try to separate things nicely and get the uncertainties in first.

jdtsmith:
Thanks guys.  We're aiming to divide #308 up into smaller more digestible PR nuggets.  The good news is the TRF fitter is working as expected and delivering uncertainties.   I'm going to close this one for now.

PR #314, opened 2026-06-24
Add normalized BlackBody/ModifiedBlackBody + guess function improvements

BenTrana27:
Note: Per [jdtsmith](https://github.com/jdtsmith)'s preferred PR ordering, this is intended to merge after the following PRs are completed:
1. Adding TRF model type
2. Add code to interpret 'param_cov' / feature table param mapping

This PR doesn't have a technical dependency on either of the PRs stated above as it has been tested and
can merge cleanly on its own; however, holding off respects the agreed sequencing.

This PR ports BB normalization and guess improvements from PR #308 (drvdputt, orion+galaxy_paper
branch) per the new PR plan in the "Ben Trana Summer" google doc (item 3).

**`ap_components.py`**
- `BlackBody1D`/`ModifiedBlackBody1D` now normalize against a fixed
  reference point (3µm/5000K and 17µm respectively) instead of raw
  amplitude. It provides the same physical fits, but better-behaved numbers internally.

**`model.py`**
- Removed an `np.clip(0, 1)` that was capping the dust continuum guess. It could hand the optimizer an artificial or wrong starting point.
- Line power guess now uses a real unit-aware formula (`fnu * dlambda * c / lambda**2`) instead of an unscaled approximation.
- Minor quote-style cleanup (single → double quotes) in the FWHM guess section, no behavior change whatsoever.

**Testing:** Ran an M83 example notebook. The fit converged, final spectrum matches the pre-change version, fitted values are all sane (no NaNs).

### DIFF
--- a/pahfit/fitters/ap_components.py
+++ b/pahfit/fitters/ap_components.py
@@ -9,6 +9,14 @@ from pahfit import units
 __all__ = ["BlackBody1D", "ModifiedBlackBody1D", "S07_attenuation", "att_Drude1D"]
 
 
+def bb(x, temperature):
+    return (
+        3.9728917e13  # 2 h c/µm^3 -> MJy
+        / x**3
+        / (np.exp(1.4387752e4 / x / temperature) - 1.0)
+    )  # h c/micron k K
+
+
 class BlackBody1D(Fittable1DModel):
     """
     A blackbody component.
@@ -20,15 +28,12 @@ class BlackBody1D(Fittable1DModel):
     amplitude = Parameter()
     temperature = Parameter()
 
+    norm = bb(3, 5000)
+
     @staticmethod
     def evaluate(x, amplitude, temperature):
         """ """
-        return (
-            amplitude
-            * 3.9728917e13 # 2 h c/µm^3 -> MJy
-            / x**3 
-            / (np.exp(1.4387752e4 / x / temperature) - 1.0)  # h c/micron k K
-        )
+        return amplitude * bb(x, temperature) / BlackBody1D.norm
 
 
 class ModifiedBlackBody1D(BlackBody1D):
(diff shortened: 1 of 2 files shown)


jdtsmith:
@drvdputt can you comment on the reasoning for normalizing all BBs at 3µm and then again at 17µm? Was this a convergence thing?

drvdputt:
The MBB is normalized twice simply because it reuses the stellar BB. I implemented the 17 um normalization first, and later realized that I also needed to normalize the stellar component, for a which a normalization at 3 um makes more sense given the typical 5000 K temperature. 

I chose a different normalization for stellar/dust components to have intuitive numbers, relative to what you see on the spectrum graph: amplitude at 3 um for stellar, amplitude at 17 um for dust. 

The reason for having a normalization like this at all, had to do with numerical precision issues (not machine precision, but rather the convergence criterion "eps = 10e-10"). My findings were documented here #296. I found that without normalization, the typical tau for the MBB has a high dynamic range. E.g. for the M101 example, tau = 1e-4 at 300 K or 1e+3 at 35 K, which is already 7 orders of magnitude, or a power law ~ (temperature / 100K)**(-8). So if we include 400 or 600 K, the fitter may stop moving along that dimension prematurely, as typical steps in tau fall below the "eps" value.

Normalizing to a specific wavelength was a simple ad-hoc choice to get "reasonable" numbers for all components. It may be possible to work with a single scaling factor for all MBB models (a unit change), to avoid the very small numbers. This will blow up the low-temperature tau to large values though, so some testing will be needed to see if this is a problem.

pahfit/fitters/ap_components.py
```diff
@@ -20,15 +28,12 @@ class BlackBody1D(Fittable1DModel):
     amplitude = Parameter()
     temperature = Parameter()
 
+    norm = bb(3, 5000)
```
jdtsmith, line 31: Normalizing twice shouldn't do anything.  I think this can be dropped (@BenTrana27 ?).

pahfit/model.py
```diff
@@ -231,7 +231,7 @@ def dust_continuum_guess(row):
             bb = ModifiedBlackBody1D(1, temp)
             flux_ref = np.median(flux[(lam > lam_ref - 0.2) & (lam < lam_ref + 0.2)])
             amp_guess = flux_ref / bb(lam_ref)
-            return np.clip(amp_guess / nbb, 0, 1.)
+            return amp_guess / nbb
```
jdtsmith: Should at least be clipped to be >0.

BenTrana27:
PR #314 fixes:

`ap_components.py` — ModifiedBlackBody1D was normalizing twice (once via BlackBody1D.evaluate(), once via its own 17µm ratio). Now calls the raw `bb()` function instead, so it only normalizes once.

`model.py` — Restored a lower-bound clip `(np.clip(amp_guess / nbb, 0, None))` on the dust continuum guess, so it can't go negative, while keeping the upper-bound cap removed.

jdtsmith:
Great work, thanks!

Issue #301, opened 2024-11-30
APFitter using deprecated optimization module


jdtsmith:
[Note from](https://github.com/astropy/astropy/blob/01777d1ef1f1a492e7a690ff3d88cf180e058303/docs/modeling/fitting.rst#notes-on-non-linear-fitting) `astropy.modeling`: 

> Note that the [:class:`~astropy.modeling.fitting.LevMarLSQFitter`](https://github.com/PAHFIT/pahfit/issues/new#id13) fitter, which uses the Levenberg-Marquardt algorithm via the scipy legacy function scipy.optimize.leastsq, is no longer recommended. This fitter supports parameter bounds via an unsophisticated min/max condition whereby during each step of the fitting, parameters that are out of bounds are simply reset to the min or max of the bounds. This can cause parameters to "stick" to one of the bounds if during the fitting process the parameter gets close to the bound. If the models you are fitting make use of bounds, you should make use of one of the other fitters...

drvdputt:
While working on fitting the Orion Bar spectra, I found myself switching between LevMarLSQFitter and TRFLSQFitter a lot, and over time I've noticed some peculiarities about each one. Today, I decided to do some testing using the classic pack + the m101 spectrum. 

There is one big advantage to using the TRFLSQFitter: _it actually returns a covariance matrix_

Now for some performance notes.

### General 
Both methods can fail when the uncertainties are unreasonable, when the fit parameters are not well normalized. Those aspects should always be checked first before concluding anything about the fitters.

### LevMarLSQFitter

- Supports bounds, but has the "sticking" problem cited above. For most use cases the effect is minor.
- I once tried setting a cap on the dust_continuum tau, and this actually prevented the fit from converging, even though it was "high enough" for the spectrum. It worked fine when the upper limit was infinity. This is one example of the "sticking".
- Less sensitive to the initial guess, seems more robust against "tough" regions in the parameter space (for example, getting away from attenuation = 0, see note about TRFLSQFitter below)

### TRFLSQFitter

- While it's supposed to be the "best" option for our use case, I find it a lot harder to get good fits with it.
- It's behavior with bounds is arguably better, as the test with the capped dust_continuum actually fit fine here.
- VERY sensitive to "bad" initial guess values. The fitter just stops (xtol or ftol condition) while improvements are clearly possible judging by eye. There must be some assumptions about parameter order-of-magnitudes going on.
- Fails to fit even the M101 classic test.

[...]

PR #317, opened 2026-06-26
Add selectable astropy fitting methods

hcha9:
Now officially implemented "lm" and "trf" fitter methods.
(Based on Dries' PR #308)
e.g.,
model.guess(spec)
model.fit(spec, maxiter=10000, use_instrument_fwhm=False, method="lm")
model.fit(spec, maxiter=100000, use_instrument_fwhm=False, method="trf")
_ = model.plot(spec, scalefac_resid=1, label_lines=False, alpha=0.2)

### DIFF
--- a/pahfit/fitters/ap_fitter.py
+++ b/pahfit/fitters/ap_fitter.py
@@ -234,13 +234,13 @@ class APFitter(Fitter):
         Parameters
         ----------
         lam : array
-            Rest-frame wavelengths in micron.
+            Rest frame wavelengths in micron.
 
         flux : array
-            Rest-frame flux in PAHFIT internal units.
+            Rest frame flux in internal units.
 
         unc : array
-            Uncertainty on the rest-frame flux.
+            Uncertainty on rest frame flux. Same units as flux.
 
         maxiter : int
             Maximum number of fitting iterations or function
@@ -252,13 +252,9 @@ class APFitter(Fitter):
             ``TRFLSQFitter``. If None, the first available method,
             currently ``"lm"``, is used.
         """
-        weights = 1.0 / unc
+        w = 1 / unc
 
-        mask = (
-            np.isfinite(lam)
-            & np.isfinite(flux)
-            & np.isfinite(weights))
-
+        mask = (np.isfinite(lam) & np.isfinite(flux) & np.isfinite(w))
         method_str = self.methods[0] if method is None else method
 
         if method_str not in self.methods:
@@ -277,16 +273,14 @@ class APFitter(Fitter):
             self.model,
             lam[mask],
             flux[mask],
-            weights=weights[mask],
+            w=w[mask],
             maxiter=maxiter,
-            epsilon=1e-10,
-            acc=1e-10)
+            epsilon=1e-7,
+            acc=1e-7)
 
         self.model = fitted_model
         self.fit_info = fitter.fit_info
         self.message = fitter.fit_info["message"]
-
-
 
     def get_result(self, component_name):
         """Retrieve results from astropy model component.
--- a/pahfit/fitters/ap_fitter.py
+++ b/pahfit/fitters/ap_fitter.py
@@ -8,7 +8,7 @@ from .ap_components import (
     PowerDrude1D,
     PowerGaussian1D,
 )
-from astropy.modeling.fitting import LevMarLSQFitter
+from astropy.modeling.fitting import LevMarLSQFitter, TRFLSQFitter
 import numpy as np
 
 
@@ -69,6 +69,8 @@ class APFitter(Fitter):
         self.feature_types = {}
         self.model = None
         self.message = None
+        self.fit_info = None
+        self.methods = ["lm", "trf"]
 
     def finalize(self):
         """Sum the registered components into one CompoundModel.
@@ -222,60 +224,69 @@ class APFitter(Fitter):
         """
         return self.model(lam)
 
-    def fit(self, lam, flux, unc, maxiter=10000):
-        """Fit the internal model using the astropy fitter.
+    def fit_methods_available(self):
+        """Return the fitting methods available for this backend."""
+        return self.methods
 
-        The fitter class is unit agnostic, and deal with the numbers the
-        Model tells it to deal with. Internal renormalizations could be
-        good to consider, as long as any values are converted back to
-        the original system before returning them. In practice, the
-        input spectrum is expected to be in internal units, and orrected
-        for redshift (models operate in the rest frame).
-
-        After the fit, the results can be retrieved via get_result().
-
-        Retrieval of uncertainties and fit details is yet to be
-        implemented.
-
-        CAVEAT: flux unit (flux) is still ambiguous, since it can be
-        flux density or intensity, according to the options defined in
-        pahfit.units. After the fit, the return units of "power" in
-        get_results depend on the given spectrum (they will be flux unit
-        times wavelength unit).
+    def fit(self, lam, flux, unc, maxiter=10000, method=None):
+        """Fit the internal model using an Astropy fitter.
 
         Parameters
         ----------
         lam : array
-            Rest frame wavelengths in micron
+            Rest-frame wavelengths in micron.
 
         flux : array
-            Rest frame flux in internal units.
+            Rest-frame flux in PAHFIT internal units.
 
         unc : array
-            Uncertainty on rest frame flux. Same units as flux.
+            Uncertainty on the rest-frame flux.
 
+        maxiter : int
+            Maximum number of fitting iterations or function
+            evaluations.
+
+        method : {"lm", "trf"} or None
+            Astropy fitting method. ``"lm"`` uses
+            ``LevMarLSQFitter`` and ``"trf"`` uses
+            ``TRFLSQFitter``. If None, the first available method,
+            currently ``"lm"``, is used.
         """
-        # clean, because astropy does not like nan
-        w = 1 / unc
+        weights = 1.0 / unc
 
-        # make sure there are no zero uncertainties either
-        mask = np.isfinite(lam) & np.isfinite(flux) & np.isfinite(w)
+        mask = (
+            np.isfinite(lam)
+            & np.isfinite(flux)
+            & np.isfinite(weights))
 
-        self.fit_info = []
+        method_str = self.methods[0] if method is None else method
 
-        fit = LevMarLSQFitter(calc_uncertainties=True)
-        astropy_result = fit(
+        if method_str not in self.methods:
+            raise PAHFITModelError(
+                f"Selected method {method} not available "
+                "for APFitter backend.")
+
+        if method_str == "lm":
+            fitter_cls = LevMarLSQFitter
+        else:
+            fitter_cls = TRFLSQFitter
+
+        fitter = fitter_cls(calc_uncertainties=True)
+
+        fitted_model = fitter(
             self.model,
             lam[mask],
             flux[mask],
-            weights=w[mask],
+            weights=weights[mask],
             maxiter=maxiter,
             epsilon=1e-10,
-            acc=1e-10,
-        )
-        self.fit_info = fit.fit_info
-        self.model = astropy_result
-        self.message = fit.fit_info["message"]
+            acc=1e-10)
+
+        self.model = fitted_model
+        self.fit_info = fitter.fit_info
+        self.message = fitter.fit_info["message"]
+
+
 
     def get_result(self, component_name):
         """Retrieve results from astropy model component.
--- a/pahfit/fitters/ap_fitter.py
+++ b/pahfit/fitters/ap_fitter.py
@@ -220,7 +220,7 @@ class APFitter(Fitter):
         Returns
         -------
         flux : array
-            Rest frame flux in internal units
+            Rest frame flux in internal units.
         """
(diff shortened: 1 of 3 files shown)


pahfit/fitters/ap_fitter.py
```diff
@@ -137,18 +137,26 @@ def add_feature_starlight(self, name, temperature, tau):
...
+            extinction law for Rv=5.5.
+
         """
         self.feature_types[name] = "dust_continuum"
         kwargs = self._astropy_model_kwargs(
-            name, ["temperature", "amplitude"], [temperature, tau]
-        )
-        self._add_component(ModifiedBlackBody1D, **kwargs)
+            name, ["temperature", "amplitude"], [temperature, tau])
+        ap_class = ModifiedBlackBody1D
```
jdtsmith: Is this temporary variable necessary?

pahfit/fitters/ap_fitter.py
```diff
@@ -254,6 +264,10 @@ def fit(self, lam, flux, unc, maxiter=10000):
         unc : array
             Uncertainty on rest frame flux. Same units as flux.
 
+        method : str
+            Fit method. Available options: "lm" will use
+            LevMarLSQFitter. "trf" will use TRFLSQFitter.
```
jdtsmith: We should mention the default is `lm`.  Maybe it should be `trf` though?

pahfit/fitters/ap_fitter.py
```diff
@@ -263,19 +277,52 @@ def fit(self, lam, flux, unc, maxiter=10000):
...
+            PAHFITModelError(
+                f"Selected method {method} not available for APFitter backend."
+            )
+
+        fitter_call_kwargs = dict(acc=1e-10)
+        if method_str == "lm":
+            fitter_cls = LevMarLSQFitter
+        elif method_str == "trf":
+            fitter_cls = TRFLSQFitter
+            fitter_call_kwargs["acc"] = 1e-10
```
jdtsmith: I'm not convinced we need to increase the accuracy like this.  @drvdputt can you comment?  I think the default is 1e-7?

pahfit/fitters/fitter.py
```diff
@@ -2,54 +2,48 @@
 
 
 class Fitter(ABC):
-    """Abstract base class for internal Fitter API.
+    """Abstract base class for interal Fitter API.
```
jdtsmith: Oops, typo

pahfit/fitters/fitter.py
```diff
@@ -2,54 +2,48 @@
...
        fitting framework are hidden behind the respective subclass.
-
-    2. Fit the model to the spectrum without any additional
-       assumptions.  The Fitter will fit the given data using the
-       given model without needing to be aware of redshift, units, or
-       other instrumental effects.
-    
+    2. Fit the model to the spectrum without any additional assumptions.
+       The Fitter will fit the given data using the given model without
+       thinking about redshift, units, instrumental effects.
```
jdtsmith: Keep spacing preserved below.

pahfit/fitters/fitter.py
```diff
@@ -2,54 +2,48 @@
...
-    that the function signatures are the same between different Fitter
-    implementations, so that uniform logic can be implemented outside
-    of the Fitter (in practice, this is a loop over the Features table
-    implemented in :class:`pahfit.model.Model`).
+    for certain standard PAHFIT quantities (in practice, these are the
+    values stored in the Features table). The abstract Fitter class
+    ensure that the function signatures are the same between different
+    Fitter implementations, so that only a single logic has to be
+    implemented to up the Fitter (in practice, this is a loop over the
+    Features table implemented in Model).
```
jdtsmith: Not sure what this means here.  Maybe:

> The abstract Fitter class ensures that the function signatures are the same between different
    Fitter implementations, so that the :class:`pahfit.fitter.Fitter` setup can remain generic (in practice, this is a loop over the  Features table implemented in :class:`pahfit.model.Model`).

pahfit/fitters/ap_fitter.py
drvdputt: This comes from the "special modified blackbody" implementation. (as is model="default").

pahfit/fitters/ap_fitter.py
drvdputt: Default is epsilon 1e-10, acc 1e-10. I think I left this in here just for some experiments. Here #301 I did make a note that I had to increase acc. Not sure if this is still true if the unit/precision/convergence problems are resolved.

pahfit/model.py
```diff
@@ -8,12 +8,11 @@
 from scipy import interpolate, integrate
 
 from pahfit import units
-from pahfit.features.util import bounded_is_fixed, bounded_is_disabled
+from pahfit.features.util import bounded_is_fixed, bounded_is_missing
```
drvdputt: I ran the tests for this branch, and `bounded_is_missing` does not exist. It is still called `bounded_is_disabled`.

drvdputt (commented):
Good to see some work getting done to get these features in one by one. 

Though, we could probably keep the changes to the plotting code for a separate one? Unless they are necessary to make the plots work after these changes.

pahfit/model.py
jdtsmith: Thanks.  We need to get our CI tests back in shape too.

pahfit/fitters/ap_fitter.py
hcha9: Default = lm

pahfit/model.py
hcha9: I changed to bounded_is_disabled as your PR. It should work now!

pahfit/fitters/ap_fitter.py
```diff
@@ -263,19 +269,52 @@ def fit(self, lam, flux, unc, maxiter=10000):
 
         self.fit_info = []
 
-        fit = LevMarLSQFitter(calc_uncertainties=True)
-        astropy_result = fit(
+        # select method based on given string or pick default if None
+        method_str = self.methods[0] if method is None else method
+        if method_str not in self.methods:
+            PAHFITModelError(
```
BenTrana27: I think there is a missing `raise` here. As written, the error object gets created but immediately discarded and execution continues, eventually hitting a `NameError` on `fitter_cls` instead of showing this message. I believe it should be `raise PAHFITModelError(...)`.

jdtsmith: I agree.

pahfit/model.py
```diff
@@ -297,17 +310,14 @@ def drude_power_guess(row):
...
-                    # A structured masked array is masked if _any_ of
-                    # its elements is masked.  Table prevents setting
-                    # values in such a masked array element, so we
-                    # access the underlying array itself with .data
-                    self.features["fwhm"].data[row_index]['val'] = line_fwhm_guess(row)
-                    for b in ('min', 'max'):
-                        self.features["fwhm"].data[row_index][b] = np.nan
-                    self.features["fwhm"].data[row_index]['frozen'] = False
+                if row["fwhm"] is np.ma.masked:
+                    self.features[row_index]["fwhm"] = (
```
BenTrana27: While testing this PR, I encountered this bug while running `model.guess(spec)` in a Jupyter notebook for an M83 fit: `ValueError: could not assign tuple of length 3 to structure with 4 fields.`

After further investigation, the tuple here only has 3 elements (value, min, max) but the fwhm column now expects 4 (val, min, max, frozen). It looks like the frozen flag got left out when this was rewritten. 

Adding `False` as a 4th element should fix it.

pahfit/fitters/ap_fitter.py
hcha9: Good point. I will revise this today. Thanks!

pahfit/model.py
hcha9: Fixed!

pahfit/fitters/ap_fitter.py
```diff
@@ -137,18 +138,26 @@ def add_feature_starlight(self, name, temperature, tau):
...
         Analogous. Temperature and tau are used as temperature and
         amplitude
 
+        Parameters
+        ----------
+        model : str (available values: 'default' and 'special')
+            Keyword to activate alternate continuum shapes. 'default' is
+            a modified blackbody with a lambda**-2 factor. With
+            model='special', the modification factor is the MW average
+            extinction law for Rv=5.5.
```
jdtsmith: Let's separate out all the "special" blackbody stuff from this PR.

pahfit/fitters/ap_fitter.py
```diff
@@ -222,8 +230,81 @@ def evaluate(self, lam):
         """
         return self.model(lam)
 
-    def fit(self, lam, flux, unc, maxiter=10000):
-        """Fit the internal model using the astropy fitter.
+
+    def _free_parameter_state(self):
+        free_param_names = []
```
jdtsmith: Needs a doc-string.

pahfit/fitters/ap_fitter.py
```diff
@@ -254,28 +337,149 @@ def fit(self, lam, flux, unc, maxiter=10000):
...
+            using the inverse norms of the Jacobian columns.
+
+        acc : float
+            Relative error desired in the approximate solution. This is
+            passed as acc to the Astropy LM fitter and as xtol to scipy
+            least_squares for TRF.
+
+        epsilon : float
+            Step_size control for numerical Jacobian estimation. The TRF
+            path follows Astropy's convention and passes diff_step=npsqrt(epsilon)
```
jdtsmith: Typo

BenTrana27:
I wanted to leave a summary here for reference. 

I tested the `x_scale='jac'` fix on M83, and `trf` executed on its own now matches `lm` and `lm→trf` almost exactly. Chi² came out within about 0.02% of each other, so the scaling fix is working as intended. While I was at it, I also tried tuning a few of the fit settings to see if anything could be squeezed out further:

- Dropping `maxiter` from 100,000 or 10,000 down to 1,000 didn't change the result at all — the fit always converges around 111 evaluations, so the higher ceiling was never actually being used.
- Changing `epsilon` from the default `1e-07` to `1e-04` cut the number of evaluations roughly in half (111 → ~46) and the fit time by about 60% (55s → 20s), with only a tiny change in chi² (~0.01%). It might be worth considering as a new default if fit speed matters, especially for batch processing multiple regions.
- `acc` didn't seem to make much difference in the range I tried (1e-10 to 1e-05).

jdtsmith:
> I wanted to leave a summary here for reference.
> 
> I tested the `x_scale='jac'` fix on M83, and `trf` executed on its own now matches `lm` and `lm→trf` almost exactly. Chi² came out within about 0.02% of each other, so the scaling fix is working as intended. While I was at it, I also tried tuning a few of the fit settings to see if anything could be squeezed out further:
> 
> * Dropping `maxiter` from 100,000 or 10,000 down to 1,000 didn't change the result at all — the fit always converges around 111 evaluations, so the higher ceiling was never actually being used.
> * Changing `epsilon` from the default `1e-07` to `1e-04` cut the number of evaluations roughly in half (111 → ~46) and the fit time by about 60% (55s → 20s), with only a tiny change in chi² (~0.01%). It might be worth considering as a new default if fit speed matters, especially for batch processing multiple regions.
> * `acc` didn't seem to make much difference in the range I tried (1e-10 to 1e-05).

This is great news!  Did you compare the actual number of interaction (FEV's) consumed by each fitter?  Can you please open this as a separate PR?

BenTrana27:
> This is great news! Did you compare the actual number of interaction (FEV's) consumed by each fitter? Can you please open this as a separate PR?

Yes! lm took 246 function evaluations, trf (baseline settings) took 111, and trf with epsilon=1e-4 took only 46. So trf is meaningfully more efficient than lm even before tuning, and roughly 5x more efficient than lm with the epsilon change, while landing on essentially the same chi².

I'm happy to open a PR for the epsilon default change. Since #317 isn't merged yet, should I base it on top of @hcha9 's branch (so it includes his trf fix) and open it as its own PR against dev, or wait until #317 merges first and open mine after, against a clean dev?

jdtsmith:
> > This is great news! Did you compare the actual number of interaction (FEV's) consumed by each fitter? Can you please open this as a separate PR?
> 
> Yes! lm took 246 function evaluations, trf (baseline settings) took 111, and trf with epsilon=1e-4 took only 46. So trf is meaningfully more efficient than lm even before tuning, and roughly 5x more efficient than lm with the epsilon change, while landing on essentially the same chi².
> 
> I'm happy to open a PR for the epsilon default change. Since #317 isn't merged yet, should I base it on top of @hcha9 's branch (so it includes his trf fix) and open it as its own PR against dev, or wait until #317 merges first and open mine after, against a clean dev?

Let's wait until #317 is merged.

pahfit/fitters/ap_fitter.py
```diff
@@ -218,64 +220,69 @@ def evaluate(self, lam):
...
+
+        fitted_model = fitter(
             self.model,
             lam[mask],
             flux[mask],
-            weights=w[mask],
+            w=w[mask],
             maxiter=maxiter,
-            epsilon=1e-10,
-            acc=1e-10,
```
jdtsmith, line 274: Save this for another small PR.

jdtsmith:
Thanks.

Issue #258, opened 2023-03-17
New Science Packs after v2.5


jdtsmith:
Now that the dev branch has merged, we need to integrate some of the improvements we've made to the science packs.  

I'm proposing a couple of new science packs:

1. `pdr.yaml` — including the observed PDR lines from the PDRs4All project.
2. `gal.yaml` — checkpointing @ThomasSYLai 's PAH33, H20 and other attenuation sources, and extending `classic.yaml` down to 2.5µm. 

The idea is that these will evolve and improve quickly as we and others use and develop these packs.  Ideas welcome.

PR #319, opened 2026-07-13
PDR pack

drvdputt:
We talked about including some more modern science packs with PAHFIT. Adding my PDR pack will be a good start. The parameters are already published across two papers (Van De Putte et al. 2025 and 2026), but I've wanted to publish the actual YAML file for a while now.

Note that this one only contains dust_feature entries. Needs to be combined with a line list and continuum model. Can be done by copy pasting for now, but eventually, combining the contents of partial science packs could be done automatically (#298).

So have a look at this one, and let me know if we would want it to be formatted differently.

### DIFF
--- a/pahfit/packs/science/pdr_pack_v2.yaml
+++ b/pahfit/packs/science/pdr_pack_v2.yaml
@@ -1,0 +1,334 @@
+#####################################
+#####################################
+# PDR pack v2. Used for Van De Putte et al. 2026A&A...710A..60V. There is no
+# official PDR pack v1; its parameters are given by the table in Van De Putte
+# et al. 2025A&A...701A.111V. For v2, 16-18 um was updated using NGC7023 data.
+#####################################
+#####################################
+
+#################
+# Dust Features #
+#################
+
+# 3.2-3.6
+# -------
+
+PAH_3.2:
+  kind: dust_feature
+  wavelength: 3.243
+  fwhm: 0.018
+
+PAH_3.3_cmp:
+  kind: dust_feature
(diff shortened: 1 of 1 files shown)


drvdputt:
As noted in #258 , we also want to have my Orion Bar gas line pack, probably. That file is not as well-polished though; I added many ad-hoc lines (without clear identification) to help mask out the numerous small features.

jdtsmith:
Thanks Dries! We will give this a close look.  Would you prefer the "multi-file" science pack format we discussed in London?  I think a "basic" all inclusive `pdr.yaml` (dust features, lines, continuum) with "add-on" packs for lines that may appear only in some spectra might be a good approach.  Thoughts?

drvdputt:
Suggesting a default continuum for PDRs would be useful. 

For the line pack that I used for NGC7023, I did not do much testing; I mainly used them as a list of things to mask out: After constructing the PAHFIT model, I loop over all lines, mask them out in the spectrum, and remove them from the features table. Would need to do some fits with the lines included to see if the values I put in there are functional; I expect there may be some duplicates or "fake" lines that I used to mark minor artifacts.

The Iris nebula line list may be a good baseline for PDRs, as it does not have strong emission from ionized regions. So a good starting point would be to include the lines listed by Misselt et al. 2025. We could then think about an add-on to include lines that are seen in the bright ionized layers, like in the Orion Bar.

We also need to decide what to do with the attenuation model. I would say, activate it by default. But in my recent paper, I disabled it (as fitting always resulted in 0 attenuation).

Issue #28, opened 2018-10-23
Use astropy.units?


karllark:
We could use astropy.units to specify the units of the spectra.  Then we could ensure they are in the units we expect.  It may even be transparent through the use of the astropy.models as long as we setup things correctly.  Worth thinking about.

jdtsmith:
Units are a bit of a pain.  Wavelength is straightforward.  But right now you have two choices for flux input in original PAHFIT: MJy/sr (in which case all integrated strengths/uncertanties/etc. are in W/m^2/sr), or "some other f_nu unit", in which case integrated values are [Unit]-Hz.  Would we want to support non-f_nu units?  Any reason to?  I don't know if astropy.units can handle composite unit systems like f_lambda vs. f_nu.

karllark:
Yes, astropy.units can handle a wide range of units.  Basically, as long as the user specifies the wavelength and spectral units, we can use astropy.units to convert our preferred units.  And convert back to the user specified units for output.  And we could also have default units if they are not specified (probably with a warning stating what defaults are  being assumed).

jdtsmith:
Interesting.  I suspect we'd prefer everything in f_nu units internally, so we can integrate f_nu dnu wherever it's needed.  Even with user-specified units, we'd still have to make choices for output units, since many outputs are (weighted) integrals over flux density.   E.g. if that give erg/s/cm^2/angstrom, we still need to provide some composite flux unit output (e.g. erg/s/cm^2).

jdtsmith:
One other niggling issue with the LM optimization PAHFIT currently does is it doesn't like very small numbers.  So depending on your units choice, the output can differ.  I often recommend people to normalize the spectrum to do the fit in numbers near unity, then rescale.  Hoping we can avoid that, or perhaps it won't be necessary.  But a good TEST would be to give PAHFIT the same exact spectrum in a variety of units, and see if it produces the same output (after unit conversion).

jdtsmith:
We discussed this issue on the call today.  One interesting question is what JWST pipeline output spectra will look like, in terms of their units attachment.  Another issue is that specifying input spectral units (e.g. MJy/sr) is not sufficient to specify an output integrated power unit (e.g. W/m^2/sr). We could either make that choice for the user, or have a default they could (rarely) modify.

jdtsmith:
Should we re-open this?

karllark:
Probably should reopen.  Have no memory of why I closed it.

karllark:
[...]

PR #321, opened 2026-07-16
Dual-unit support: flux density + surface brightness + getting units out of features table

BenTrana27:
## Summary

Adds dual-unit support to PAHFIT (#28): accepts both flux-density (Jy, mJy,
and now wavelength-based units like erg/s/cm²/Å) and surface-brightness
(MJy/sr) input, converts correctly internally, and labels output units
based on what was actually used.

## What's done

- Detects flux vs. surface brightness input (`is_surface_brightness()`)
- Accepts flux spectra instead of rejecting them; removed the old fake
  solid-angle workaround
- Fitting math and initial guesses use the correct flux/intensity factors
  based on input type, via a shared `working_units()` helper (replacing
  duplicated logic across `_convert_spec_data`, `guess()`, and the
  `Power*` fitting components)
- `PowerDrude1D`/`PowerGaussian1D` now compute their amplitude factor
  generically at evaluate time, instead of two separate hardcoded
  constants per class
- Wavelength-based flux units (e.g. `erg/s/cm²/Å`) now supported via
  `astropy`'s `spectral_density` equivalency
- `power` column labels itself correctly after fitting
- Added `get_quantity()` to pull a table value with its unit attached,
  including graceful handling of masked values (`None`, not a fake `0`)
  and dimensionless columns like `tau` (previously crashed)

## Testing

Tested end-to-end on real M83 data, both input types. Surface-brightness
results unchanged from before this PR. Flux results cross-checked against
surface-brightness results via implied solid angle which was consistent to ~2-3%
across two features. `get_quantity()` tested across every feature and
every numeric column in a real fitted table (265 checks, 0 errors).
This testing caught and fixed the `tau` crash above.

## Open questions

- Uncertainty support via `get_quantity()` is out of scope for this PR. It will be
  tracked as a separate future PR, pending the covariance matrix work.

### DIFF
--- a/pahfit/fitters/ap_components.py
+++ b/pahfit/fitters/ap_components.py
@@ -176,6 +176,10 @@ class PowerDrude1D(Fittable1DModel):
         .to(units.intensity)
         .value
     )
+
+    # ASSUMPTION (needs Dr. Smith's verification): mirrors
+    # intensity_amplitude_factor above, substituting flux_power/
+    # flux_density. Dimensionally valid; not yet physically confirmed.
     flux_amplitude_factor = (
         (2 * units.flux_power * units.wavelength / (constants.c * np.pi))
         .to(units.flux_density)
@@ -279,6 +283,10 @@ class PowerGaussian1D(Fittable1DModel):
         .to(units.intensity)
         .value
     )
+
+    # ASSUMPTION (needs Dr. Smith's verification): mirrors
+    # intensity_amplitude_factor above, substituting flux_power/
+    # flux_density. Dimensionally valid; not yet physically confirmed.
     flux_amplitude_factor = (
         (
             units.flux_power
--- a/pahfit/fitters/ap_components.py
+++ b/pahfit/fitters/ap_components.py
@@ -159,22 +159,26 @@ class PowerDrude1D(Fittable1DModel):
     are the internal pahfit units defined in pahfit.units, and
     precalculate a conversion factor.
 
-    TODO: We need to check if the flux is 'intensity' or 'flux_density',
-    and assume the power parameter has 'intensity_power' or
-    'flux_density_power' units respectively. For now, only intensity is
-    supported.
-
+    Both intensity (surface brightness, MJy/sr) and flux density (mJy)
+    inputs are supported. The appropriate amplitude factor is selected
+    at evaluate time based on which unit track is active.
     """
-
     power = Parameter(min=0.0)
     x_0 = Parameter(min=0.0)
     fwhm = Parameter(default=1, min=0.0)
-
+    
     # constant factors in the equation to convert power to amplitude of
-    # the profile.
+    # the profile. Two versions: one for surface brightness input
+    # (intensity_power -> intensity) and one for flux density input
+    # (flux_power -> flux_density).
     intensity_amplitude_factor = (
         (2 * units.intensity_power * units.wavelength / (constants.c * np.pi))
         .to(units.intensity)
+        .value
+    )
+    flux_amplitude_factor = (
+        (2 * units.flux_power * units.wavelength / (constants.c * np.pi))
+        .to(units.flux_density)
         .value
     )
 
@@ -221,7 +225,8 @@ class PowerDrude1D(Fittable1DModel):
         # factor = (2 * unit(power) * unit(wavelength) / (pi * c)).to(unit(intensity))
 
         g = fwhm / x_0
-        b = power * x_0 / g * self.intensity_amplitude_factor
+        factor = self.flux_amplitude_factor if getattr(self, 'is_flux', False) else self.intensity_amplitude_factor
+        b = power * x_0 / g * factor
         return b * g**2 / ((x / x_0 - x_0 / x) ** 2 + g**2)
 
 
@@ -256,6 +261,9 @@ class PowerGaussian1D(Fittable1DModel):
     So the constant factor we can set is
     (unit(power) * unit(wavelength)**2 / (c * unit(wavelength) * sqrt(2 pi))).to(intensity)
 
+    Both intensity (surface brightness, MJy/sr) and flux density (mJy)
+    inputs are supported. The appropriate amplitude factor is selected
+    at evaluate time based on which unit track is active.
     """
 
     power = Parameter(min=0.0)
@@ -271,6 +279,15 @@ class PowerGaussian1D(Fittable1DModel):
         .to(units.intensity)
         .value
     )
+    flux_amplitude_factor = (
+        (
+            units.flux_power
+            * (units.wavelength) ** 2
+            / (constants.c * units.wavelength * np.sqrt(2 * np.pi))
+        )
+        .to(units.flux_density)
+        .value
+    )
 
     def evaluate(self, x, power, mean, stddev):
         """
@@ -278,6 +295,7 @@ class PowerGaussian1D(Fittable1DModel):
 
         See class description for equations and unit notes."""
 
-        # amplitude in intensity units
-        Anu = power * mean**2 / stddev * self.intensity_amplitude_factor
+        # amplitude in intensity or flux density units
+        factor = self.flux_amplitude_factor if getattr(self, 'is_flux', False) else self.intensity_amplitude_factor
+        Anu = power * mean**2 / stddev * factor
         return Anu * np.exp(-0.5 * np.square((x - mean) / stddev))
--- a/pahfit/fitters/ap_components.py
+++ b/pahfit/fitters/ap_components.py
@@ -170,11 +170,6 @@ class PowerDrude1D(Fittable1DModel):
     def _amplitude_factor(self):
         """Conversion factor from power to profile amplitude, computed
         for whichever unit track (flux or surface brightness) is active.
-
-        # ASSUMPTION (needs Dr. Smith's verification): this generic
-        # formula was verified to reproduce the exact same numeric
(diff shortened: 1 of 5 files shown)


jdtsmith:
Great, thanks.  You can convert $f_\lambda \leftrightarrow f_\nu$ using [spectral equivalencies](https://docs.astropy.org/en/stable/units/equivalencies.html#:~:text=1.%20Hz%3E-,Spectral,-Units).

pahfit/model.py
```diff
@@ -429,6 +430,17 @@ def _ingest_fit_result_to_features(self):
...
+        # ASSUMPTION (flag for Dr. Smith): power values are only
+        # calculated correctly for surface-brightness input right now.
+        # If the input was flux, warn instead of silently mislabeling.
+        if not units.is_surface_brightness(self.features.meta["user_unit"]["flux"]):
+            warnings.warn(
+                "Input spectrum is flux density (not surface brightness). "
+                "Fitted 'power' values are not yet dimensionally correct "
+                "for this case — this is a known limitation, not a bug in "
+                "your data. See PowerDrude1D/PowerGaussianSum1D docstring."
+            )
```
jdtsmith: Not sure what's meant here.  We have an integrated power unit for flux, it's in W/m^2.  So we just need to convert to that...

BenTrana27: I believe this comment is on an earlier commit. That warning is now gone in the final version. The final version does exactly this: `flux_power` (1e-22 W/m^2) was added in `units.py`, and after a fit, the `power` column gets relabeled to it when the input is flux.

pahfit/fitters/ap_components.py
```diff
@@ -166,24 +159,26 @@ class PowerDrude1D(Fittable1DModel):
...
     fwhm = Parameter(default=1, min=0.0)
 
-    # constant factors in the equation to convert power to amplitude of
-    # the profile.
-    intensity_amplitude_factor = (
-        (2 * units.intensity_power * units.wavelength / (constants.c * np.pi))
-        .to(units.intensity)
-        .value
-    )
+    def _amplitude_factor(self):
```
BenTrana27: Do this once in the `__init__` function.

BenTrana27: The model will want to decide ONCE.

pahfit/fitters/ap_components.py
```diff
@@ -263,28 +258,38 @@ class PowerGaussian1D(Fittable1DModel):
...
     power = Parameter(min=0.0)
     mean = Parameter()
     stddev = Parameter(default=1, min=0.0)
 
-    intensity_amplitude_factor = (
-        (
-            units.intensity_power
-            * (units.wavelength) ** 2
-            / (constants.c * units.wavelength * np.sqrt(2 * np.pi))
+    def _amplitude_factor(self):
```
BenTrana27: Parallel this once in the __init__ function. 
The model will want to decide ONCE.

pahfit/fitters/ap_components.py
```diff
@@ -263,28 +261,41 @@ class PowerGaussian1D(Fittable1DModel):
...
     def evaluate(self, x, power, mean, stddev):
         """
         Evaluate F_nu(lambda) given the power.
 
         See class description for equations and unit notes."""
 
-        # amplitude in intensity units
-        Anu = power * mean**2 / stddev * self.intensity_amplitude_factor
+        # amplitude in intensity or flux density units
+        Anu = power * mean**2 / stddev * self._amplitude_factor()
```
jdtsmith: Seems a bit elaborate.  Why not just something like:

```python
def __init__(self, *args, **kwargs):
    self._amplitude_factor = self._compute_amplitude_factor()
    super().__init__(*args, **kwargs)
```

jdtsmith: (to initialize the amplitude once).

BenTrana27: Good suggestion! Moving it to happen once instead of every `evaluate()` call makes sense. I believe one small tweak needed though: `is_flux` isn't set until after `__init__` runs as it gets assigned later, in
`ap_fitter.py`'s `fit()`, in this loop:

    c.is_flux = self.is_flux

So I can compute it right there instead, in that same loop, at the point `is_flux` is actually known. That way `evaluate()` can just read `self._amplitude_factor` directly, like you suggested. I will start on this change now.

BenTrana27:
**Status update:** 

I have merged the latest upstream/dev into this branch to pick up #317 (TRF method selection), since both PRs touched ap_fitter.py and model.py in overlapping spots. Conflicts are resolved -- both my `is_flux/_amplitude_factor` setup and the `TRF method-selection` logic are now combined correctly in both files.

While testing the merged result against real M83 data, found what looks like a pre-existing bug in #317's fit() call: it passes `w=w[mask]` to astropy's fitter, but astropy expects `weights=`, causing a TypeError ("got an unexpected keyword argument 'w'"). This isn't something my changes introduced. It's in the code as merged from #317,  but it's
blocking a full re-test of the merged branch. I will plan to fix it locally on my branch soon.

I am planning to do more thorough testing before pushing the merge. I will follow up once I've confirmed everything works end-to-end, including the flux/TRF combination specifically.

BenTrana27:
> **Status update:**
> 
> I have merged the latest upstream/dev into this branch to pick up #317 (TRF method selection), since both PRs touched ap_fitter.py and model.py in overlapping spots. Conflicts are resolved -- both my `is_flux/_amplitude_factor` setup and the `TRF method-selection` logic are now combined correctly in both files.
> 
> While testing the merged result against real M83 data, found what looks like a pre-existing bug in #317's fit() call: it passes `w=w[mask]` to astropy's fitter, but astropy expects `weights=`, causing a TypeError ("got an unexpected keyword argument 'w'"). This isn't something my changes introduced. It's in the code as merged from #317, but it's blocking a full re-test of the merged branch. I will plan to fix it locally on my branch soon.
> 
> I am planning to do more thorough testing before pushing the merge. I will follow up once I've confirmed everything works end-to-end, including the flux/TRF combination specifically.

### **Follow-up:** 

I have fixed the `w=/weights=` bug locally and pushed it to this PR. All units functionality was re-tested post-merge and works correctly on the default "lm" method.

One thing that I flagged, but not fixed here: comparing "lm" vs "trf" on the same M83 data:
- Surface brightness: lm = 22854.25, trf = 22791.78 (1e-10 W/sr/m^2),
  ~0.27% apart -- normal optimizer variance.
- Flux: lm = 24878318531.99, trf = 50480686.99 (1e-22 W/m^2),
  ~493x apart.

**The flux/trf gap is far outside normal variance and looks like TRF-specific behavior rather than a units issue**, since surface brightness doesn't show it. I believe it may be out of scope for this PR. I will plan to test this issue and track it separately as well as provide any updates when testing.

Otherwise, this PR is ready for final review.

pahfit/fitters/ap_components.py
jdtsmith: Good point.  I suppose you could use the same model to fit a spectrum with different units flavoring. So probably need to check on each fit and recompute (once) if it's changed (maybe start by initializing is_flux to None).

BenTrana27: Great idea! I can switch `is_flux` to initialize as None instead of False, and only recompute `_amplitude_factor` when it's actually changed since the last set. That way a reused model correctly recomputes when flavor changes, and a brand-new model still always computes it at least once on the first fit. I will make this change now.

BenTrana27: > Good point. I suppose you could use the same model to fit a spectrum with different units flavoring. So probably need to check on each fit and recompute (once) if it's changed (maybe start by initializing is_flux to None).

Two fixes are committed and pushed: initializing `is_flux` to `None` so it always computes at least once on a fresh fit, and correcting the power column's unit label so it resets properly (not just sets to flux_power) when reusing a Model across different unit flavors.

I confirmed both are working correctly: fitted values are correct, and the unit labels are correct, whether going flux-to-surface-brightness or surface-brightness-to-flux, even when reusing the same Model object across fits. This was tested with the full features table, not just a single value. e.g. [NeII] power reads 24878318531.99 (1e-22 W/m^2) after a flux fit, then correctly relabels to 22872.22 (1e-10 W/(sr m2)) after fitting surface-brightness data with the same model object.

However, the "skip recompute if unchanged" part does not actually work. I checked by counting calls to `_compute_amplitude_factor()` across two fits with the same flavor. It ran 23 times on the first fit and another 23 on the second (46 total), when the second fit should have added zero if the skip were working. It turns out every `fit()` call builds brand new `PowerDrude1D` and `PowerGaussian1D` objects from scratch, so there's never a previous `is_flux` to compare against. It always recomputes no matter what. Nothing gives wrong answers because of this, but the optimization isn't happening.

I believe a fix to this issue would mean tracking `is_flux` somewhere that persists across fits, like on Model or APFitter itself, rather than on the components that get rebuilt each time. This feels like a bigger change, so I wanted to check with you before progressing further. I'm happy to look at it more in-depth if you'd like, or we can leave it as a follow-up for later.

pahfit/fitters/ap_fitter.py
```diff
@@ -275,7 +288,7 @@ def fit(self, lam, flux, unc, maxiter=10000, method=None):
             self.model,
             lam[mask],
             flux[mask],
-            w=w[mask],
+            weights=w[mask],
```
BenTrana27: I may end up removing this from this PR and consider making a new PR for this issue.

